PR #3, opened 2017-08-16
Some quality-of-life enhancements

astronouth7303:
* Make an actual `Sequence`
* Support keys with [`.__index__()`](https://docs.python.org/3/reference/datamodel.html#object.__index__)
* Be more vague about allowed types


astronouth7303:
Actually, I think making it a real `Sequence` means this now works?

```python
x, y = Vector2(4, 2)
assert x == 4 and y == 2
```

PR #22, opened 2017-10-13
Added main to execute test

royopa:


pathunstrom:
Project is going to be using pytest going forward, I'd rather not further tie unittest into the project.

royopa:
Ok, thanks for the information

PR #32, opened 2017-10-13
Create test_vector2_equality.py

Maybeking:
Testing equality and inequality

### DIFF
--- a/tests/test_vector2_equality.py
+++ b/tests/test_vector2_equality.py
@@ -1,0 +1,13 @@
+import ppb_vector
+
+def test_equal():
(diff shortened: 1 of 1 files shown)


pathunstrom:
Have you run this through a linter? Your logic is fine, and while I'd like a few more test cases for each function, I would accept this if the code followed [PEP8](http://legacy.python.org/dev/peps/pep-0008/) standards.

The specific issues:
* two space indents when there should be 4.
* there should be two lines of white space above each function.

Make those fixes and I'll be able to merge this.

PR #33, opened 2017-10-13
test_vector2_length.py

Maowse:
Added additional tests for the length function

### DIFF
--- a/tests/test_vector2_length.py
+++ b/tests/test_vector2_length.py
@@ -1,6 +1,7 @@
 import ppb_vector
 import pytest
 
+
 @pytest.mark.parametrize("x, y, expected", [
     (6, 8, 10),
     (8, 6, 10),
@@ -8,7 +9,6 @@ import pytest
     (-6, -8, 10),
     (1, 2, 2.23606797749979)
 ])
-
 def test_length(x, y, expected):
     vector = ppb_vector.Vector2(x, y)
     assert vector.length == expected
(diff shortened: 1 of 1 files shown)


pathunstrom:
This is a great set of tests!

Could I get you to take out the unittest and use test functions? We'll be using pytest going forward.

Bonus would be to parameterize them, but that work can be done going forward.

Maowse:
I removed unittest and used parametrized testing from pytest, please let me know if anything needs changed (or if I did something incorrectly!)

tests/test_vector2_length.py
```diff
@@ -0,0 +1,14 @@
+import ppb_vector
+import pytest
+
```
pathunstrom: Need one more line of white space here.

tests/test_vector2_length.py
```diff
@@ -0,0 +1,14 @@
...
+@pytest.mark.parametrize("x, y, expected", [
+    (6, 8, 10),
+    (8, 6, 10),
+    (0, 0, 0),
+    (-6, -8, 10),
+    (1, 2, 2.23606797749979)
+])
+
```
pathunstrom: And this white space should be removed a decorator should end on the line immediately preceding the function definition.

pathunstrom:
I would normally do:

    import pytest

    import ppb_vector

Because of a basic rule of grouping imports into three things: standard library imports, third party imports, local imports. In this case, ppb_vector counts as "local" because that's what we're testing.

I don't think this is a hard requirement and the way you did it is actually fine!

Issue #9, opened 2017-10-12
Test dot product


pathunstrom:
Test dot product using the `*` operator between `Vector2`s and vector-like objects.

himanshukgp:
I want to take up this issue.

astronouth7303:
This seems pretty well covered at this point?

PR #37, opened 2017-10-13
create test_vector2_dot_product.py

himanshukgp:


pathunstrom:
In reference to #9

tests/test_vector2_dot_product.py
```diff
@@ -0,0 +1,11 @@
+import ppb_vector
+
```
pathunstrom: This should be two lines of white space. Please refer to [PEP8](https://www.python.org/dev/peps/pep-0008/) for style guide.

PR #42, opened 2017-10-22
minor change in readme.md

ashishkg0022:
minor change in readme.md


pathunstrom:
Thanks for the suggestion, but because both functions require a parameter, making it look like a call in that description creates what looks like pasteable code that will fail.

I don't want samples in code blocks (including spans) to be error raising.

PR #51, opened 2017-12-23
Modified the reflect function to allow reflections even if the angle between the vector and surface normal is acute

avik-pal:
The formula hold only if the angle between the surface normal and the ray is obtuse. An additional check is provided to check if the angle is obtuse and if not the ray is reversed to allow reflection.

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -85,9 +85,6 @@ class Vector2(Sequence):
         yield self.x
         yield self.y
 
-    def __neg__(self):
-        return self.rotate(180)
-
     def rotate(self, degrees):
         r = radians(degrees)
         r_cos = cos(r)
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -85,6 +85,9 @@ class Vector2(Sequence):
         yield self.x
         yield self.y
 
+    def __neg__(self):
+        return self.rotate(180)
+
     def rotate(self, degrees):
         r = radians(degrees)
         r_cos = cos(r)
@@ -114,6 +117,7 @@ class Vector2(Sequence):
         """
         if not (0.99999 < surface_normal.length < 1.00001):
             raise ValueError("Reflection requires a normalized vector.")
-        if self * surface_normal > 0:
-            self = -self
-        return self - (2 * (self * surface_normal) * surface_normal)
+        vec_new = self
+        if self * surface_normal>0:
+            vec_new = self.rotate(180)
+        return vec_new - (2 * (vec_new * surface_normal) * surface_normal)
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -9,7 +9,7 @@ class Vector2(Sequence):
         self.x = x
         self.y = y
         self.length = hypot(x, y)
-    
+
     def __len__(self):
         return 2
 
@@ -114,4 +114,6 @@ class Vector2(Sequence):
         """
         if not (0.99999 < surface_normal.length < 1.00001):
             raise ValueError("Reflection requires a normalized vector.")
+        if self * surface_normal > 0:
+            self = -self
         return self - (2 * (self * surface_normal) * surface_normal)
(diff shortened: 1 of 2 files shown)


pathunstrom:
Your changes broke the tests.

```
============================= test session starts ==============================
platform darwin -- Python 3.6.3, pytest-3.3.1, py-1.5.2, pluggy-0.6.0
rootdir: /Users/piper/src/pathunstrom/ppb-vector, inifile:
plugins: cov-2.5.1
collected 52 items                                                             

tests/test_vector2.py .                                                  [  1%]
tests/test_vector2_addition.py .............                             [ 26%]
tests/test_vector2_length.py .....                                       [ 36%]
tests/test_vector2_member_access.py ...                                  [ 42%]
tests/test_vector2_reflect.py ....F                                      [ 51%]
tests/test_vector2_rotate.py .......                                     [ 65%]
tests/test_vector2_scalar_multiplication.py .....                        [ 75%]
tests/test_vector2_substraction.py .............                         [100%]

---------- coverage: platform darwin, python 3.6.3-final-0 -----------
Name                     Stmts   Miss  Cover   Missing
------------------------------------------------------
ppb_vector/__init__.py       1      0   100%
ppb_vector/vector2.py       90     20    78%   25, 35, 43, 58, 65-67, 76, 79-82, 97, 100-102, 105-109, 116
------------------------------------------------------
TOTAL                       91     20    78%


=================================== FAILURES ===================================
________ test_reflect[initial_vector4-surface_normal4-expected_vector4] ________

initial_vector = Vector2(-1, -1), surface_normal = Vector2(-1, 0)
expected_vector = Vector2(-1, 1)

    @pytest.mark.parametrize("initial_vector, surface_normal, expected_vector", reflect_data)
    def test_reflect(initial_vector, surface_normal, expected_vector):
>       assert initial_vector.reflect(surface_normal) == expected_vector

tests/test_vector2_reflect.py:15: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Vector2(-1, -1), surface_normal = Vector2(-1, 0)

    def reflect(self, surface_normal: 'Vector2') -> 'Vector2':
        """
            Calculate the reflection of the vector against a given surface normal
            """
        if not (0.99999 < surface_normal.length < 1.00001):
            raise ValueError("Reflection requires a normalized vector.")
        if self * surface_normal > 0:
>           self = -self
E           TypeError: bad operand type for unary -: 'Vector2'

ppb_vector/vector2.py:118: TypeError
===================== 1 failed, 51 passed in 0.20 seconds ======================
```

I did remove your code change and can confirm your new test fails, which means we need to resolve how best to handle this.

ppb_vector/vector2.py
```diff
@@ -114,4 +114,6 @@ def reflect(self, surface_normal: 'Vector2') -> 'Vector2':
         """
         if not (0.99999 < surface_normal.length < 1.00001):
             raise ValueError("Reflection requires a normalized vector.")
+        if self * surface_normal > 0:
```
pathunstrom: We're calculating the dot-product twice here. Here and on line 119.

Would love if we can only calculate once.

ppb_vector/vector2.py
```diff
@@ -114,4 +114,6 @@ def reflect(self, surface_normal: 'Vector2') -> 'Vector2':
...
+            self = -self
```
pathunstrom: This is also a thing I would rather we avoid: self should always refer to the class being operated on, reassigning to self is a smell.

Can we add a separate variable for this?

ppb_vector/vector2.py
pathunstrom: I was wrong about this. Please ignore.

ppb_vector/vector2.py
```diff
@@ -85,6 +85,9 @@ def __iter__(self):
         yield self.x
         yield self.y
 
+    def __neg__(self):
```
pathunstrom: Please remove. One major segment of work per PR.

PR #56, opened 2018-10-10
Allow Vector-likes in all operations

astronouth7303:
Make all operations work with vector-likes.

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -28,7 +28,7 @@ def _mkvector(value):
     elif isinstance(value, dict) and 'x' in value and 'y' in value and len(value) == 2:
         return Vector2(value['x'], value['y'])
     else:
-        raise ValueError("Cannot use {} as a vector-like".format(value))
+        raise ValueError(f"Cannot use {value} as a vector-like")
 
 
 class Vector2(Sequence):
@@ -93,7 +93,7 @@ class Vector2(Sequence):
             raise TypeError
 
     def __repr__(self) -> str:
-        return "{}({}, {})".format(type(self).__name__, self.x, self.y)
+        return f"{type(self).__name__}({self.x}, {self.y})"
 
     def __eq__(self, other: VectorLike) -> bool:
         if is_vector_like(other):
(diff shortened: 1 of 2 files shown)


pathunstrom:
I like where we're going with this.

Suggesting we put most of this under a class method `.from(value: Union[Sequence, Mapping[str, numbers]])` that just checks length then tries to instantiate and raises a Value error if it fails. Most of the logic is clear: Must be length 2, and look like a vector (in the case of maps), and then we can get a meaningful error everywhere, also provides a utility for people doing things programatically.

astronouth7303:
Standard python is that conversion should be in the constructor, ie `Vector2({Vector-like})`, but otherwise yes.

pathunstrom:
I only suggest a class method because the constructor is `Vector2(x, y)` and I want to keep that as a general rule. Class method as an alternative constructor is pretty well established pattern (see [`int.from_bytes`](https://docs.python.org/3/library/stdtypes.html#int.from_bytes)).

ppb_vector/vector2.py
```diff
@@ -67,16 +84,18 @@ def __getitem__(self, item):
             raise TypeError
 
     def __repr__(self):
-        return "Vector2({}, {})".format(self.x, self.y)
+        return "{}({}, {})".format(type(self).__name__, self.x, self.y)
```
pathunstrom: Since we're modifying this, lets switch to F-strings.

ppb_vector/vector2.py
```diff
@@ -115,6 +134,7 @@ def reflect(self, surface_normal: 'Vector2') -> 'Vector2':
         """
         Calculate the reflection of the vector against a given surface normal
         """
+        surface_normal = _mkvector(surface_normal)
```
pathunstrom: If we do this, we need to change the type hint.

On that front, we should probably build a union or type var for this.

astronouth7303:
Addressed everything.

Not super happy with the name `.convert()`, but I couldn't come up with a better one.

astronouth7303:
`from` is a keyword; can't be used as method names.

PR #57, opened 2018-10-10
Add cross-product and angle

nbraud:


nbraud:
I used this in [mutant-games](https://gitlab.com/astronouth7303/mutant-games/merge_requests/1), but didn't yet add tests.  Feel free to push directly commits to the feature branch to do so.

nbraud:
PS: I added a test for `angle()`, which uncovered a bug, and I fixed it.
This probably still needs a test for `a ^ b`.

README.md
```diff
@@ -87,6 +87,14 @@ Multiply a `Vector2` by another `Vector2` to get the dot product.
     >>> Vector2(45, 60).length
     75.0
 
+### Cross-product and angle
```
pathunstrom: These should have separate headings, and a line of explanation for each.

ppb_vector/vector2.py
```diff
@@ -46,6 +46,10 @@ def __rmul__(self, other):
         if isinstance(other, Number):
             return Vector2(self.x * other, self.y * other)
 
+    def __xor__(self, other):
+        assert(isinstance(other, Vector2))
```
pathunstrom: Please don't do this. Assertion errors in running code aren't meaningful for the end user.

tests/test_vector2_rotate.py
```diff
@@ -11,9 +11,14 @@
...
 ]
 
+def isclose(x, y, epsilon = 6.5e-5):
+    d = (x - y) % 360
+    return (d < epsilon) or (d > 360 - epsilon)
+
 @pytest.mark.parametrize('input, degrees, expected', data)
 def test_multiple_rotations(input, degrees, expected):
     assert input.rotate(degrees) == expected
+    assert isclose(input.angle(expected), degrees)
```
pathunstrom: 👍

ppb_vector/vector2.py
astronouth7303: This should be redone with my PR.

tests/test_vector2_rotate.py
nbraud: I had to write a custom `isclose` (rather than `math.isclose`) to deal with arithmetic modulo 360°, but I'm also pretty happy with it  :)

nbraud:
@pathunstrom Thanks for the feedback, this should all be handled now.

FYI, you can automatically squash my fixup commits with `git rebase -i --autosquash master` before merging (or I can do that once you have reviewed the new changes)

PR #60, opened 2018-10-11
[WIP] Add Hypothesis

### LINKED ISSUE #59: Add property-based testing
I think it would make sense, in many cases, to run tests not against hardcoded inputs (and expected outputs), but against random inputs (and properties/equations that should hold).
For instance, `x.angle(x.rotate(alpha)) ≃ alpha` for any vector `x` and angle `alpha`.

Given that `ppb-vector` is a math library, and written in an effects-free style, it seems to be a great fit for property-based testing, and I can recommend the [Hypothesis](https://hypothesis.works/) property-based testing library.

It would probably make sense to first add some automation to track line- or path-coverage of the testsuite, ideally ran automatically in CI, but this isn't a strong blocker either.

astronouth7303:
Add property-based testing to search for bugs.

Fixes #59 
Requires #58

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /.cache/
 /*.egg-info/
 __pycache__/
+.hypothesis/
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+/.cache/
+/*.egg-info/
+__pycache__/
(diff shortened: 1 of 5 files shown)


astronouth7303:
It did find a bug in `.reflect()`

PR #62, opened 2018-10-13
Hypothesis: Initial support

nbraud:
Supersedes #60, we are redoing it to split off the various part.

This only contains the basic scaffolding to be able to write Hypothesis tests, no new tests or functional changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /.cache/
 /*.egg-info/
 __pycache__/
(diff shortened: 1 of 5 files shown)


astronouth7303:
Sorry, this PR also kinda got away a bit.

astronouth7303:
Addressed everything.

Not super happy with the name `.convert()`, but I couldn't come up with a better one.

PR #63, opened 2018-10-13
Hypothesis: Add test for reflect()

nbraud:
~Based off #62~

The test failed without 248491f, because `tests.utils.units` returned not-quite-normalized vectors, because `Vector2.rotate()` rounded things off.  ~(You can check, adding `.normalize()` to the returned value in `units` “fixes” it.)~


nbraud:
This should be good to go once `fix-rotation` is merged.

PR #64, opened 2018-10-13
Hypothesis: Add tests for rotate()

nbraud:
Based off #62 
Thanks to @astronouth7303 for writing the initial version of the test.

@pathunstrom If you confirm that the issue uncovered by Hypothesis is a bug, I can add commits fixing it to this branch.


nbraud:
@astronouth7303 @pathunstrom Actually, Hypothesis finds some annoying numerical instabilities even without the rounding (due to the way we compute `rotate` and `angle`).
Would it be reasonable to provide a more stable implementation in this PR, i.e. add tests that expose the problem (already done) then fix it?

astronouth7303:
I would say create a separate PR for it. If you discover more bugs, this PR could become large.

nbraud:
@astronouth7303 So, ship as-is and send another PR to fix the bug exposed by the test?

astronouth7303:
In the other order, yes.

nbraud:
@astronouth7303 Weird, but OK.

nbraud:
#67 contains the first 2 commits, which fix the bug exposed by Hypothesis.

nbraud:
Had to pull in #69, because 5c6c1a9 wasn't a correct test of linearity (it didn't test multiplication by a scalar)

tests/test_vector2_rotate.py
```diff
@@ -24,6 +27,54 @@ def test_for_exception():
...
+        stepwise = stepwise.rotate(angle)
+    note(f"Step-wise: {stepwise}")
+
+    assert fellswoop.isclose(stepwise)
+    assert math.isclose(fellswoop.length, initial.length)
+
+
+@given(
+    a=vectors(max_magnitude=1e150), b=vectors(),
+    l=st.floats(min_value=1e150, max_value=1e150),
```
astronouth7303: 1e150 to 1e150 seems wrong.

astronouth7303:
The blocker for this right now is that I've found a clear failure of the linearity test.

nbraud:
@astronouth7303 I rebased and cleaned up the history a bit (the resulting diff is equivalent, though).

Also, I noticed that the comment on the example you provide says “Rotation must not be an multiple of 90deg”, while you set `angle=45`.

astronouth7303:
Yes, in order to trigger the test case, the angle must not be axis-aligned (0, 90, 180, 270, etc). So the demo data uses 45deg as the sample.

I listed out the rules of how to trigger the bug, as best as I could discover them.

nbraud:
Oh, yes, I'm very good at reading and missed the “not” there.  Sorry

nbraud:
Finally fixed  <3

astronouth7303:
Sweet! Is there anything else that needs to happen?

nbraud:
@astronouth7303 Not for this branch?  Tests pass :D

PR #67, opened 2018-10-17
Fix rounding bug in Vector2.rotation

nbraud:
See #64

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -1,6 +1,6 @@
 import typing
 import collections
-from math import acos, cos, degrees, hypot, radians, sin
+from math import acos, cos, degrees, hypot, isclose, radians, sin
 from numbers import Real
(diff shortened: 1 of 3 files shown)


nbraud:
This should also make the CI go green on #63, as it was failing due to the `rotate()` bug.

ppb_vector/vector2.py
```diff
@@ -163,8 +163,10 @@ def rotate(self, degrees: Real) -> 'Vector2':
         r = radians(degrees)
         r_cos = cos(r)
         r_sin = sin(r)
-        x = round(self.x * r_cos - self.y * r_sin, 5)
-        y = round(self.x * r_sin + self.y * r_cos, 5)
+        assert isclose(r_cos * r_cos + r_sin * r_sin, 1)
```
astronouth7303: Please move this into a test.

PR #70, opened 2018-10-18
Add property-based tests for cross-products

nbraud:
PR targetting the branch for #65 

- [x] Add linearity and anti-symmetry test
- [x] Document that cross-products may overflow on large vectors
- [x] Restrict the generated vectors to avoid overflow (pulled in #69)
- [ ] Fix the exposed numerical instabilities in cross-products.

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -158,7 +158,10 @@ class Vector2:
 
     def __xor__(self: VectorOrSub, other: VectorLike) -> Realish:
         """
-        Computes the magnitude of the cross product
+        Computes the magnitude of the cross product.
+
+        May overflow on the product of two large vectors
+        (operands length >= 10¹⁵⁰) and produce NaN.
         """
         other = Vector2.convert(other)
         return self.x * other.y - self.y * other.x
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -158,7 +158,7 @@ class Vector2:
 
     def __xor__(self: VectorOrSub, other: VectorLike) -> Realish:
         """
-        Computes the magnitude of the cross product.
+        Computes the z-coordinate of the cross product.
 
         May overflow on the product of two large vectors
         (operands length >= 10**150) and produce NaN.
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -161,7 +161,7 @@ class Vector2:
         Computes the magnitude of the cross product.
 
         May overflow on the product of two large vectors
-        (operands length >= 10¹⁵⁰) and produce NaN.
+        (operands length >= 10**150) and produce NaN.
         """
         other = Vector2.convert(other)
         return self.x * other.y - self.y * other.x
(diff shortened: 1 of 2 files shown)


ppb_vector/vector2.py
```diff
@@ -84,7 +84,10 @@ def __rmul__(self, other: VectorLike) -> 'Vector2':
 
     def __xor__(self, other: VectorLike) -> Real:
         """
-        Computes the magnitude of the cross product
+        Computes the magnitude of the cross product.
+
+        May overflow on the product of two large vectors
+        (operands length >= 10¹⁵⁰) and produce NaN.
```
astronouth7303: Can we keep the unicode out of the source? I know most editors should be handle UTF-8 at this point, but.... yeah.

nbraud:
PS: #65 was merged without this, so it should be re-targetted against master

astronouth7303:
There is a failure of the linearity test, as well. I assume this is the instability you've referred to before?

nbraud:
@astronouth7303 @pathunstrom I was poking at making the cross-product more numerically stable, when I realised that there's nothing that uses it anymore since #68 was merged.

I kinda feel like removing the feature outright might be sensible, since providing a well-behaved cross-product is pretty hard.  What do you think?

astronouth7303:
This seems reasonable, since this only exists because you needed it for some algorithm.

Is there something external that you'd want it for? Intersections? Sidedness?

nbraud:
It can be used for sidedness, yes.  If someone decides they need it, it would be pretty easy to bring it back.

nbraud:
PS: Just pushed some local changes I had, for the record, but if you confirm we can remove the feature, I would rather do that.

nbraud:
@astronouth7303 Mmmh, cross-product is also nice to check if vectors are aligned.
(`x` and `y` are aligned iff `x ^ y ≃ 0`)

nbraud:
@pathunstrom The dot product is a bit more annoying to use for that purpose, IMO (you have to normalize & such).  OTOH, I just realised this makes no sense, we have a perfectly fine `Vector2.angle` method, which is probably the better way to check whether things are aligned.

nbraud:
Replying to @astronouth7303's comments on #80 (I think it would be better to keep the conversation here) :
> What algorithms depend on (or significantly benefit from) cross product, even ones not included in this library?

Nothing in the library use cross-product, and I can't think of anything I would want in a game that benefits from it; for general geometry, cross-product is pretty convenient for computing the area of triangles or parallelograms between 2 vectors (area_plg = 2*area_triangle = abs(a ^ b)).

> Can we characterize the instability? Is it something that only appears at extremes or are there perfectly reasonable vectors that have major problems?

Yes, the instability occurs when there are near-parallel vectors, IIRC.
It keeps occurring even if I crank up the relative tolerance to 0.01 (i.e. tolerate 10% error) and with small (but not ludicruously-small) vectors.

astronouth7303:
That's a pretty significant instability (10% will probably have noticeable effects in a game) for normal vectors. I think that qualifies cross product as a foot-gun and it should probably be removed on that premise.

astronouth7303:
@pathunstrom your vote?

pathunstrom:
I'm all for keeping foot guns out of the ppb name space.

nbraud:
:heart:

PR #71, opened 2018-10-20
Fix return types

astronouth7303:
* Actual algorithm for mixed-type operations
* Fix more annotations
* Tests to verify return types

There is an argument to be made for Vector/Vector operations to return the left type.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,3 +36,4 @@ install:
   
 script:
   - pytest
+  - mypy ppb_vector tests
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -12,12 +12,14 @@ __all__ = 'Vector2',
 VectorOrSub = typing.TypeVar('VectorOrSub', bound='Vector2')
 
 # Anything convertable to a Vector, including lists, tuples, and dicts
-VectorLike = typing.TypeVar('VectorLike', typing.Union[
+VectorLike = typing.Union[
     VectorOrSub,
     typing.List[Real],  # TODO: Length 2
     typing.Tuple[Real, Real],
     typing.Dict[str, Real],  # TODO: Length 2, keys 'x', 'y'
-])
+]
+
+Realish = typing.Union[Real, float, int]
 
 
 def is_vector_like(value: typing.Any) -> bool:
@@ -70,9 +72,12 @@ def _find_lowest_vector(left: typing.Type, right: typing.Type) -> typing.Type:
         return _find_lowest_type(left, right)
 
 
-class Vector2(Sequence):
+class Vector2:
+    x: Realish
+    y: Realish
+    length: float
 
-    def __init__(self: VectorOrSub, x: Real, y: Real):
+    def __init__(self: VectorOrSub, x: Realish, y: Realish):
         self.x = x
         self.y = y
         self.length = hypot(x, y)
@@ -103,19 +108,34 @@ class Vector2(Sequence):
         rtype = _find_lowest_vector(type(other), type(self))
         return rtype(self.x - other.x, self.y - other.y)
 
-    def __mul__(self: VectorOrSub, other: typing.Union[VectorLike, Real]) -> typing.Union[VectorOrSub, Real]:
+    def dot(self: VectorOrSub, other: VectorLike) -> Realish:
+        """
+        Return the dot product of two vectors.
+        """
+        other = _mkvector(other)
+        return self.x * other.x + self.y * other.y
+
+    def scale_by(self: VectorOrSub, other: Realish) -> VectorOrSub:
+        """
+        Scale by the given amount.
+        """
+        return type(self)(self.x * other, self.y * other)
+
+    def __mul__(self: VectorOrSub, other: typing.Union[VectorLike, Realish]) -> typing.Union[VectorOrSub, Realish]:
+        """
+        Performs a dot product or scale based on other.
+        """
         if is_vector_like(other):
             try:
-                other = _mkvector(other)
+                return self.dot(other)
             except ValueError:
                 return NotImplemented
-            return self.x * other.x + self.y * other.y
         elif isinstance(other, Real):
-            return type(self)(self.x * other, self.y * other)
+            return self.scale_by(other)
         else:
             return NotImplemented
 
-    def __rmul__(self: VectorOrSub, other: typing.Union[VectorLike, Real]) -> typing.Union[VectorOrSub, Real]:
+    def __rmul__(self: VectorOrSub, other: typing.Union[VectorLike, Realish]) -> typing.Union[VectorOrSub, Realish]:
         return self.__mul__(other)
 
     def __xor__(self: VectorOrSub, other: VectorLike) -> Real:
@@ -125,7 +145,7 @@ class Vector2(Sequence):
         other = _mkvector(other)
         return self.x * other.y - self.y * other.x
 
-    def __getitem__(self: VectorOrSub, item: typing.Union[str, int]) -> Real:
+    def __getitem__(self: VectorOrSub, item: typing.Union[str, int]) -> Realish:
         if hasattr(item, '__index__'):
             item = item.__index__()
         if isinstance(item, str):
@@ -162,14 +182,14 @@ class Vector2(Sequence):
         else:
             return True
 
-    def __iter__(self: VectorOrSub) -> typing.Iterator[Real]:
+    def __iter__(self: VectorOrSub) -> typing.Iterator[Realish]:
         yield self.x
         yield self.y
 
     def __neg__(self: VectorOrSub) -> VectorOrSub:
-        return self * -1
+        return self.scale_by(-1)
 
-    def angle(self: VectorOrSub, other: VectorLike) -> Real:
+    def angle(self: VectorOrSub, other: VectorLike) -> Realish:
         other = _mkvector(other, castto=Vector2)
 
         rv = degrees( atan2(other.x, -other.y) - atan2(self.x, -self.y) )
@@ -182,7 +202,7 @@ class Vector2(Sequence):
 
         return rv
 
-    def isclose(self: VectorOrSub, other: VectorLike, *, rel_tol: Real=1e-06, abs_tol: Real=1e-3) -> bool:
+    def isclose(self: VectorOrSub, other: VectorLike, *, rel_tol: Realish=1e-06, abs_tol: Realish=1e-3) -> bool:
         """
         Determine whether two vectors are close in value.
 
@@ -198,13 +218,15 @@ class Vector2(Sequence):
         For the values to be considered close, the difference between them
         must be smaller than at least one of the tolerances.
         """
+        other = _mkvector(other, castto=Vector2)
         diff = (self - other).length
         return (
-            diff < rel_tol * max(self.length, other.length) or 
+            diff < rel_tol * self.length or
+            diff < rel_tol * other.length or 
             diff < abs_tol
         )
 
-    def rotate(self: VectorOrSub, degrees: Real) -> VectorOrSub:
+    def rotate(self: VectorOrSub, degrees: Realish) -> VectorOrSub:
         r = radians(degrees)
         r_cos = cos(r)
         r_sin = sin(r)
@@ -216,17 +238,22 @@ class Vector2(Sequence):
     def normalize(self: VectorOrSub) -> VectorOrSub:
         return self.scale(1)
 
-    def truncate(self: VectorOrSub, max_length: Real) -> VectorOrSub:
+    def truncate(self: VectorOrSub, max_length: Realish) -> VectorOrSub:
         if self.length > max_length:
             return self.scale(max_length)
         return self
 
-    def scale(self: VectorOrSub, length: Real) -> VectorOrSub:
+    def scale_to(self: VectorOrSub, length: Real) -> VectorOrSub:
+        """
+        Scale the vector to the given length
+        """
         try:
             scale = length / self.length
         except ZeroDivisionError:
             scale = 1
         return self * scale
+
+    scale = scale_to
 
     def reflect(self: VectorOrSub, surface_normal: VectorLike) -> VectorOrSub:
         """
@@ -237,3 +264,5 @@ class Vector2(Sequence):
             raise ValueError("Reflection requires a normalized vector.")
 
         return self - (2 * (self * surface_normal) * surface_normal)
+
+Sequence.register(Vector2)
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -73,14 +73,20 @@ def _find_lowest_vector(left: typing.Type, right: typing.Type) -> typing.Type:
 
 
 class Vector2:
-    x: float
-    y: float
-    length: float
+    x: Realish
+    y: Realish
+    length: Realish
 
     def __init__(self: VectorOrSub, x: Realish, y: Realish):
-        self.x = float(x)
-        self.y = float(y)
-        self.length = hypot(x, y)
+        try:
+            self.x = x.__float__()
+        except AttributeError:
+            raise TypeError(f"{type(x).__name__} object not convertable to float")
+        try:
+            self.y = y.__float__()
+        except AttributeError:
+            raise TypeError(f"{type(y).__name__} object not convertable to float")
+        self.length = hypot(self.x, self.y)
 
     @classmethod
     def convert(cls: typing.Type[VectorOrSub], value: VectorLike) -> VectorOrSub:
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -26,21 +26,15 @@ def is_vector_like(value: typing.Any) -> bool:
     return isinstance(value, (Vector2, list, tuple, dict))
 
 
-class _fakevector(typing.NamedTuple):
-    x: float
-    y: float
-
-
-t_mkvector = typing.TypeVar('t_mkvector', 'Vector2', _fakevector, covariant=True)
-def _mkvector(value: VectorLike, *, castto: typing.Type[t_mkvector]=_fakevector) -> t_mkvector:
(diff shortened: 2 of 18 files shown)


nbraud:
@astronouth7303 I don't understand what's going on.

Is it making vector/vector operations that involve subclasses of `Vector2` return the subclass? If so, what's the motivation?

astronouth7303:
Consistency of typing without every subclass needing to wrap each operation to return the right type.

Concretely, so `ppb.vector.Vector` appears and avoiding `Vector2` showing up.

nbraud:
Oh, I didn't realise `ppb.Vector` was different from `Vector2`.

:+1: then, as it causes the bug I mentioned in mutant-game

nbraud:
The type annotations are probably wrong, though.

astronouth7303:
Subclasses of the thing are implied in all annotations, afaik.

So, they all return a `Vector2`. It might be a specialization of `Vector2`, but it'll be a `Vector2`.

nbraud:
Yes, but what I mean is that code that does a vector operation, then calls a subclass method/property on the result, would work with your change, but fail to typecheck because the typechecker thinks we return a `Vector2`

astronouth7303:
https://mypy.readthedocs.io/en/latest/kinds_of_types.html#class-types

> Any instance of a subclass is also compatible with all superclasses

So a type annotation of `Vector2` (or it's late-bound variant `'Vector2'`) implicitly includes all subclasses of `Vector2`.

nbraud:
@astronouth7303 OK, to clarify:

```python3
class V(Vector2):
    def method(self):
        pass

x = V(1, 0) + (0, 1)  # x's inferred type will be Vector2, not V
x.method()  # Should fail to typecheck
```

astronouth7303:
Oh, then yes. I think this is one of the limitations of the python type checker, since (afaik), you can't have a type variable that goes between different bits of a method's annotation.

Like you can't say "x.foo() will return whatever type x is".

pathunstrom:
That sounds like a thing that should be discussed on mypy.

tests/test_typing.py
```diff
@@ -0,0 +1,77 @@
+import pytest
+
+from ppb_vector import Vector2
+
+# List of operations that (Vector2, Vector2) -> Vector2
```
pathunstrom: This suggest that the variable names could be more clear. It took me a few moments to realize the name meant "Binary operations."

tests/test_typing.py
```diff
@@ -0,0 +1,77 @@
...
+
+@pytest.mark.parametrize('op', BINOPS)
+def test_binop_subclass(op):
+    class V1(Vector2): pass
+    class V2(V1): pass
+
+    # Normalize for reflect
+    a = op(V1(1, 2), V2(3, 4).normalize())
+    b = op(V2(1, 2), V1(3, 4).normalize())
+    assert isinstance(a, V2)
```
pathunstrom: Do you mean specifically that we're returning the subclass in all cases?

ppb_vector/vector2.py
```diff
@@ -34,6 +35,37 @@ def _mkvector(value, *, castto=_fakevector):
...
+    if len(lmro) > len(rmro):
+        return left
+    elif len(rmro) > len(lmro):
+        return right
+    else:
+        # They're equal, just arbitrarily pick one
+        return left
+
+
+def _find_lowest_vector(left: type, right: type) -> type:
```
pathunstrom: Not actionable: I don't like the amount of private functions that this is growing into. I understand where it's coming from, thus, non-actionable.

ppb_vector/vector2.py
```diff
@@ -34,6 +35,37 @@ def _mkvector(value, *, castto=_fakevector):
         raise ValueError(f"Cannot use {value} as a vector-like")
 
 
+@functools.lru_cache()
+def _find_lowest_type(left: type, right: type) -> type:
```
pathunstrom: Should probably use typing.Type here.

tests/test_typing.py
astronouth7303: Yes, because V2 is a subclass of V1, it should return the more specialized V2.

ppb_vector/vector2.py
astronouth7303: These are documented as equivalent:

> `Type[Any]` is equivalent to `Type` which in turn is equivalent to `type`, which is the root of Python’s metaclass hierarchy.

But I'll do it because metaprogramming like this can be weird.

nbraud:
@astronouth7303 It looks like it's doable: https://gist.github.com/nbraud/a592985bb700ac74894efd8243cb08ce

astronouth7303:
Everything mentioned so far should be taken care of.

astronouth7303:
With annotations getting this complex, I figured we should actually run it through mypy.

It was informative.

astronouth7303:
https://github.com/python/mypy/issues/5821 came up, and has some insights into typing.

astronouth7303:
Sorry, this PR is getting kinda large. I feel like I've touched most of the code at this point.

nbraud:
@astronouth7303 Yeah, there is likely no way to improve the annotations without touching all the methods.  Thanks a lot for doing it.

Small question about `Realish` (which I pronounce “relish” in my head, it's funny): I had also noticed that mypy complains about `float`s and `int`s not being `Real`s, but CPython seems to disagree:
```python3
> from numbers import Real

> isinstance(1, Real)
True

> isinstance(1.0, Real)
True
```

Is mypy in the wrong about `int`s and `float`s not being `Real`s?

nbraud:
Re: the PR becoming too large, I guess we can split off the actual behaviour change (in selecting the return type) and the improvement in type annotations.  What do you think?

astronouth7303:
I suppose I can. I'm not sure it helps much?

astronouth7303:
> Small question about Realish (which I pronounce “relish” in my head, it's funny): I had also noticed that mypy complains about floats and ints not being Reals, but CPython seems to disagree:

This is because mypy doesn't pickup on [runtime registered ABC subclasses](https://docs.python.org/3/library/abc.html#abc.ABCMeta.register).

So, mypy is correct in that `numbers.Real` is not in the MRO of `int` and `float`. But, if you ask if `int` is a subclass of `Real` at runtime, the answer will be yes (because it's explicitly registered as such).

nbraud:
@astronouth7303 Sounds like a bug that should be reported against mypy, then?
In any case, we should definitely keep the `Realish` union for now, but if the `mypy` devs fix it, we can in the future remove it.

astronouth7303:
That's Real Complicated, and there's been Very Large Discussions about mypy and ABCs and The Number Tower (PEP 3141).

.travis.yml
```diff
@@ -36,3 +36,4 @@ install:
   
 script:
   - pytest
+  - mypy ppb_vector tests
```
nbraud: I think you need to set `MYPYPATH` to be `${VIRTUAL_ENV}`, otherwise `mypy` will complain about being unable to resolve the `hypothesis` imports.

ppb_vector/vector2.py
```diff
@@ -1,97 +1,174 @@
...
+    y: float
+    _length: float
 
-    def __init__(self, x: Real, y: Real):
-        self.x = x
-        self.y = y
-        self.length = hypot(x, y)
+    def __init__(self, x: typing.SupportsFloat, y: typing.SupportsFloat):
+        try:
+            self.x = x.__float__()
```
nbraud: Why is this calling the dunder, rather than using `float()` ?

ppb_vector/vector2.py
```diff
@@ -1,97 +1,174 @@
...
+
+# Anything convertable to a Vector, including lists, tuples, and dicts
 VectorLike = typing.Union[
-    'Vector2',
-    typing.List[Real],  # TODO: Length 2
-    typing.Tuple[Real, Real],
-    typing.Dict[str, Real],  # TODO: Length 2, keys 'x', 'y'
+    'Vector2',  # Or subclasses, unconnected to the VectorOrSub typevar above
+    typing.List[typing.SupportsFloat],  # TODO: Length 2
+    typing.Tuple[typing.SupportsFloat, typing.SupportsFloat],
```
nbraud: Replace `List` and `Tuple` with `typing.Sequence` ?

ppb_vector/vector2.py
```diff
@@ -1,97 +1,174 @@
...
         """
-        return _mkvector(value, castto=type(cls))
+        # Use Vector2.convert() instead of type(self).convert() so that 
+        # _find_lowest_vector() can resolve things well.
+        if isinstance(value, cls):
+            return value
+        elif isinstance(value, Vector2):
+            return cls(value.x, value.y)
+        # FIXME: Allow all types of sequences
+        elif isinstance(value, (list, tuple)) and len(value) == 2:
```
nbraud: `isinstance(value, typing.Sequence)` ?

ppb_vector/vector2.py
```diff
@@ -1,97 +1,174 @@
...
+            return cls(value[0].__float__(), value[1].__float__())
+        # FIXME: Allow all types of mappings
+        elif isinstance(value, dict) and 'x' in value and 'y' in value and len(value) == 2:
```
nbraud: `isinstance(value, typing.Mapping)`?

Also, do we want to enforce `len(value) == 2`?
It might be useful to be able to build a vector from a dict like `{'x': 0, 'y': 1, 'color': 'brown'}`.

ppb_vector/vector2.py
```diff
@@ -1,97 +1,174 @@
...
+            return cls(value['x'].__float__(), value['y'].__float__())
+        else:
+            raise ValueError(f"Cannot use {value} as a vector-like")
+
+    @property
+    def length(self) -> float:
+        if not hasattr(self, '_length'):
```
nbraud: Is the reflection dance with `hasattr` to memoize the vector length really faster than recomputing it every time?

ppb_vector/vector2.py
```diff
@@ -1,97 +1,174 @@
...
 
-    def __sub__(self, other: VectorLike) -> 'Vector2':
+    def __sub__(self: VectorOrSub, other: VectorLike) -> VectorOrSub:
+        rtype = _find_lowest_vector(type(other), type(self))
         try:
-            other = _mkvector(other)
+            other = Vector2.convert(other)
         except ValueError:
             return NotImplemented
-        rtype = type(other) if isinstance(other, Vector2) else type(self)
```
nbraud: Is it a good idea to move the computation of `rtype` before the `try/except`?
It can be that `_find_lowest_vector` throws an exception in cases where we would previously have returned `NotImplemented`

ppb_vector/vector2.py
```diff
@@ -109,32 +186,32 @@ def __getitem__(self, item: typing.Union[str, int]) -> Real:
...
+    def __eq__(self: VectorOrSub, other: typing.Any) -> bool:
         if is_vector_like(other):
-            other = _mkvector(other)
+            other = Vector2.convert(other)
             return self.x == other.x and self.y == other.y
         else:
             return False
 
-    def __ne__(self, other: VectorLike) -> bool:
+    def __ne__(self: VectorOrSub, other: typing.Any) -> bool:
```
nbraud: Can't we replace the whole function with `not self == other`?

tests/test_vector2_angle.py
```diff
@@ -10,7 +10,7 @@
     (Vector2(1, 1), Vector2(-1, 0), 135),
     (Vector2(0, 1), Vector2(0, -1), 180),
     (Vector2(-1, -1), Vector2(1, 0), 135),
-    (Vector2(-1, -1), Vector2(-1, 0), 45),
+    (Vector2(-1, -1), Vector2(-1, 0), -45),
```
nbraud: Why are there unrelated changed in the angle tests?

tests/test_vector2_reflect.py
```diff
@@ -38,7 +38,7 @@ def test_reflect_prop(initial: Vector2, normal: Vector2):
     assert initial.isclose(returned)
     note(f"initial ⋅ normal: {initial * normal}")
     note(f"reflected ⋅ normal: {reflected * normal}")
-    assert isclose((initial * normal), -(reflected * normal))
+    assert isclose((initial.dot(normal)), -(reflected.dot(normal)))
```
nbraud: Why is this change needed?

ppb_vector/vector2.py
nbraud: OK, I did a quick [μbench](https://gist.github.com/nbraud/2dd0ae4904f7e97727f51722a1bba007) with `timeit`, and it looks like the speedup/slowdown of saving the result in `_length` is below the measurement noise, so I would recommend removing that code and just doing `return hypot(self.x, self.y)`.

```
$ ./dunder_bench.py
1 successive length computations
        Memoized: 0.41153700300492346
        Computed: 0.41805144600220956
        Speedup:  1.5829543758445919%
5 successive length computations
        Memoized: 1.2452478090126533
        Computed: 1.2318236450082622
        Speedup:  -1.0780315297269971%
10 successive length computations
        Memoized: 2.252152765984647
        Computed: 2.2965358549845405
        Speedup:  1.970696200996347%
25 successive length computations
        Memoized: 5.274616469978355
        Computed: 5.400348082999699
        Speedup:  2.3837110003537276%
50 successive length computations
        Memoized: 10.367006862012204
        Computed: 10.45787951600505
        Speedup:  0.8765563214376826%
100 successive length computations
        Memoized: 21.056249847984873
        Computed: 21.302848722989438
        Speedup:  1.1711433744606925%
./dunder_bench.py  81.75s user 0.00s system 99% cpu 1:21.77 total
```

nbraud:
@astronouth7303 Thanks for the link to the mypy discussions, re: the numbers tower.

ppb_vector/vector2.py
astronouth7303: Disallowing strings.

.travis.yml
astronouth7303: IS THAT WHAT"S GOING ON?????????????

astronouth7303: Except the build isn't using virtual environments.

ppb_vector/vector2.py
astronouth7303: The code only supports `list` and `tuple`, so ideally yes, but that's not what's implemented yet.

ppb_vector/vector2.py
astronouth7303: You'd want [`collections.abc.Sequence`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence)

ppb_vector/vector2.py
astronouth7303: For now, I think we want to keep exact forms to prevent mistakes.

Especially since afaik no code actually uses the dict form.

ppb_vector/vector2.py
astronouth7303: They don't throw exceptions, and they are internal functions meant for DRYing.

ppb_vector/vector2.py
astronouth7303: Probably.

tests/test_vector2_reflect.py
astronouth7303: Because I changed it before discovering `@override` and didn't feel like switching it back.

.travis.yml
astronouth7303: Yeah, this is only a thing that's happening locally.

ppb_vector/vector2.py
astronouth7303: Does computing it have any significant overhead overhead? I did this to get rid of  `_fakevector` while keeping the performance benefit of not computing the length at construction time.

tests/test_vector2_angle.py
astronouth7303: Because passing tests.

ppb_vector/vector2.py
nbraud: @astronouth7303 That's what the microbenchmark I wrote was showing: even if you access the computed length 100 times in a row, loading it from `self._length` isn't faster than recomputing it.

tests/test_vector2_angle.py
nbraud: Except that the tests are passing in `master`, so there is something wrong going on.

ppb_vector/vector2.py
nbraud: You added `Mapping` to the imports in a6161b9ea916fd893b4493fcf29d2b747ca1f360 but you forgot to do anything with it.

ppb_vector/vector2.py
nbraud: OK, sounds good <3

tests/test_vector2_reflect.py
nbraud: Yet the code right above it still uses `*`.
Please switch back to `*` for consistency (and also because it would be hella weird not to use the preferred notation in tests)

ppb_vector/vector2.py
astronouth7303: `typing.Sequence` is basically a typing-related data structure, only used to construct annotation data.

`collections.abc.Sequence` is an actual abstract class with default implementations of a pile of methods.

tests/test_vector2_angle.py
astronouth7303: You're going to make my life difficult, I know it.

ppb_vector/vector2.py
astronouth7303: I did, this comment is outdated.

ppb_vector/vector2.py
astronouth7303: I simplified the implementation of `length`.

tests/test_vector2_angle.py
astronouth7303: *OK.*

This is actually a bug mypy found.

The two tests in this file had the same name, so only the latter one was being run.

This exposed problems in the earlier test because it wasn't running before.

This can be moved to its own PR if y'all would really rather.

PR #78, opened 2018-12-06
Vector2: Remove superfluous _length attribute

nbraud:
It was used for memoizing length computations until #71.
I proved in a microbenchmark that the memoization was providing no benefit (as `math.hypot` isn't any slower than Python's reflection API, used to check if the value was already computed)


astronouth7303:
Oops thanks!

PR #80, opened 2018-12-08
Vector2: Remove cross-product

nbraud:
Cf #70 for the rationale. TL;DR:
- hard to implement in a stable and predictable way;
- initially added to implement `Vector2.angle`, which doesn't use it anymore;
- was never included in a release of ppb-vector.


nbraud:
The CI failure seems to be some `mypy` issue unrelated to this, that I can't reproduce locally.

astronouth7303:
I'm blocking this on resolving the discussion in #70.

What algorithms depend on (or significantly benefit from) cross product, even ones not included in this library? Is the instability a deal-breaker in these cases?

Can we characterize the instability? Is it something that only appears at extremes or are there perfectly reasonable vectors that have major problems?

PR #81, opened 2018-12-08
Vector2.scale_to: Add Hypothesis tests, fix exposed bugs

nbraud:
- [x] Turned test_scale_is_equivalent_to_truncate into an Hypothesis test.
- [x] Raise a ZeroDivisionError when scaling a null vector, rather than silently swallowing the error.
- [x] Raise a ValueError when scaling to a negative length.


nbraud:
The CI failure seems to be some `mypy` issue unrelated to this, that I can't reproduce locally.

Issue #75, opened 2018-11-01
Testing parameter names should be improved for readability


pathunstrom:
It took me a bit to realize `l` in https://github.com/ppb/ppb-vector/pull/64 was length, and if I can't read it, newer devs who don't have domain knowledge _definitely_ can't.

Let's avoid single character variable names everywhere we can. Exceptions include `x` and `y` inside much of the code.

nbraud:
You mean in `test_rotation_linearity`?
If so, `l` isn't being used as a length, but as a(n arbitrary) scalar to multiply `a` with.

I've tried to keep my use of letters consistent (i.e. `x`, `y`, ... for vectors, `l` for the scalar in a linearity test, ...), but clearly it's wasn't 100% consistent (that test uses `a` and `b` for vectors).

PR #85, opened 2018-12-10
tests: Clarify the property-based tests

nbraud:
@pathunstrom raised in #75 that the variable names used aren't consistent.
This ~doesn't quite fix that (yet?), but at least it~ should make it clearer what's being tested.

I chose not to document the one-liner test, as a comment containing the same code as the assertion below it would be pretty useless.

### DIFF
--- a/tests/test_vector2_angle.py
+++ b/tests/test_vector2_angle.py
@@ -50,9 +50,9 @@ def test_angle_additive(left, middle, right):
     lr = left.angle(right)
     assert angle_isclose(lm + mr, lr)
 
-@given(x=vectors(), l=floats())
-def test_angle_aligned(x: Vector2, l: float):
-    """x.angle(l * x) is 0 or 180, depending on whether l > 0"""
-    assume(l != 0)
-    y = l * x
-    assert angle_isclose(x.angle(y), 0 if l > 0 else 180)
+@given(x=vectors(), scalar=floats())
+def test_angle_aligned(x: Vector2, scalar: float):
+    """x.angle(scalar * x) is 0 or 180, depending on whether scalar > 0"""
+    assume(scalar != 0)
+    y = scalar * x
+    assert angle_isclose(x.angle(y), 0 if scalar > 0 else 180)
--- a/tests/test_vector2_angle.py
+++ b/tests/test_vector2_angle.py
@@ -28,6 +28,10 @@ def test_angle(left, right, expected):
     right=vectors(),
 )
 def test_angle_range(left, right):
+    """Vector2.angle produces values in [-180; 180] and is antisymmetric.
+
+    Antisymmetry means that left.angle(right) == - right.angle(left).
+    """
     lr = left.angle(right)
     rl = right.angle(left)
     assert -180 < lr <= 180
@@ -40,6 +44,7 @@ def test_angle_range(left, right):
     right=vectors(),
 )
 def test_angle_additive(left, middle, right):
+    """left.angle(middle) + middle.angle(right) == left.angle(right)"""
     lm = left.angle(middle)
     mr = middle.angle(right)
     lr = left.angle(right)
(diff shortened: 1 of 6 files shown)


tests/test_vector2_angle.py
```diff
@@ -29,6 +29,7 @@ def test_angle(left, right, expected):
     right=vectors(),
 )
 def test_angle_range(left, right):
+    """Vector2.angle produces values in [-180; 180] and is antisymmetric."""
```
astronouth7303: Is antisymmetric different from asymmetric?

nbraud: Yes.  Antisymmetric means (in that case) that `left.angle(right) == - right.angle(left)`: it's symmetric, but with a twist (a sign change).
I will clarify that.

astronouth7303:
I'm going to ask that @pathunstrom review this one because she's the one that originally filed the bug, and I understood the tests already.

tests/test_vector2_angle.py
nbraud: BTW, I considered changing those vector names to `x` and `y`, since `left` and `right` aren't particularly meaningful there.

tests/test_vector2_rotate.py
```diff
@@ -15,20 +15,25 @@
     (Vector2(math.pi, math.e), 67, Vector2(-1.27467, 3.95397))
 ]
 
-@pytest.mark.parametrize('input, degrees, expected', data)
-def test_multiple_rotations(input, degrees, expected):
-    assert input.rotate(degrees).isclose(expected)
-    assert angle_isclose(input.angle(expected), degrees)
+@pytest.mark.parametrize('input, angle, expected', data)
```
pathunstrom: I don't know if I'd have called this one a necessary change, but I feel like it's more clear what you're going for here.

tests/test_vector2_rotate.py
```diff
@@ -88,26 +96,27 @@ def test_rotation_stability2(initial, angles):
...
     angle=45,
 )
-def test_rotation_linearity(a, b, l, angle):
-    inner = (l * a + b).rotate(angle)
-    outer = l * a.rotate(angle) + b.rotate(angle)
-    note(f"l * a + b: {l * a + b}")
-    note(f"l * a.rotate(): {l * a.rotate(angle)}")
-    note(f"b.rotate(): {b.rotate(angle)}")
+def test_rotation_linearity(x, y, l, angle):
+    """(l*x + y).rotate is equivalent to l*x.rotate + y.rotate"""
```
pathunstrom: This helps, I feel like we could actually name `l` as something. Even as simple as `scalar` might be enough.

tests/test_vector2_angle.py
nbraud: Is dc0f688 enough clarification on antisymmetry?
Do you have an opinion on the variable name change I suggested?

astronouth7303: Names seem reasonable, or `vector_a` and `vector_b`.

Ehhh.... defining it doesn't actually feel better?

nbraud:
@pathunstrom Thanks a lot for the feedback; I 100% agree that it is an accessibility issue, I just wasn't sure what to do to make the notation more accessible, beyond making it consistent, and your comment helped a lot :purple_heart:

nbraud:
PS: Rebased to solve the merge conflict with #88

nbraud:
I just rebased to fix the latest merge conflict.

Is there any blocker on this?  I'm getting quite tired of needing to fix merge conflicts whenever any other PR touches the tests.

tests/test_vector2_angle.py
nbraud: What would you suggest then?  Having a name for the property is useful, IMO (and that's a commonly-accepted name)

astronouth7303: I meant more that "antisymmetric" doesn't mean much to non-mathematicians.

Simply adding as an additional paragraph that flipping the order flips the sign is more helpful than trying to summarize it in a single word.

nbraud: Yes, and there is a clarification right afterwards:
```
def test_angle_range(left, right):
    """Vector2.angle produces values in [-180; 180] and is antisymmetric.

    Antisymmetry means that left.angle(right) == - right.angle(left).
    """
```

astronouth7303:
@pathunstrom Is this progress or still as opaque to you?

nbraud:
Rebased again after the last merges.

nbraud:
@astronouth7303 Can this go forward?

PR #79, opened 2018-12-08
Vector2: Add vector-scalar division

nbraud:
This is notation that's both convenient, and pretty usual in vector math.
`x / λ` is equivalent (up to floating-point errors) to `(1 / λ) * x`

PR #87, opened 2018-12-11
Improve the Readme

nbraud:
- [x] Reorganised things to separate convenience accessors, math operators, and methods.
- [x] Documented various notations that weren't mentioned.
- [x] Specified the behaviour of vector negation and the range of `Vector2.angle`.
- [x] Fixed errors in the documentation of `scale` and `truncate`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,14 +117,6 @@ Multiply a `Vector2` by another `Vector2` to get the dot product.
     >>> Vector2(45, 60).length
     75.0
 
-### Cross-product
-
-Take the cross-product between two (2D) vectors.
-The result is expressed as a scalar, as it is known to lie on the z-axis.
-
-    >>> Vector(1, 0) ^ Vector(0, 1)
-    1
-
 ### Negation
 
 Negating a `Vector2` is equivalent to multiplying it by -1.
--- a/README.md
+++ b/README.md
@@ -127,6 +127,8 @@ The result is expressed as a scalar, as it is known to lie on the z-axis.
 
 ### Negation
 
+Negating a `Vector2` is equivalent to multiplying it by -1.
+
     >>> -Vector2(1, 1)
     Vector2(-1.0, -1.0)
 
--- a/README.md
+++ b/README.md
@@ -167,10 +167,24 @@ Return a vector with the same direction, and unit length.
 
 ### scale(scalar)
 
-Scale given `Vector2` to length of `scalar`.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -140,33 +156,35 @@ Compute the angle between two vectors, expressed as a scalar in degrees.
 As with `rotate()`, angles are signed, and refer to a direct coordinate system
 (i.e. positive rotations are counter-clockwise).
 
-#### normalize()
+`Vector2.angle` is guaranteed to produce an angle between -180° and 180°.
 
-Return the normalized `Vector2` for the given `Vector2`.
+### normalize()
+
+Return a vector with the same direction, and unit length.
```
pathunstrom: Love this change.

nbraud: <3

nbraud:
@pathunstrom The CI is going to fail because there's one `doctest` issue I'm unsure how to handle:

    Failed example:
        Vector2(1, 0).rotate(90)
    Expected:
        Vector2(0.0, 1.0)
    Got:
        Vector2(6.123233995736766e-17, 1.0)

Ideally, I would like `Vector2.rotate` to produce the right result, but the previous rounding behaviour introduced a bunch of quirks that were pretty undesireable.
I think it should be possible to make it happen by selecting values of `cos_r` and ` sin_r` such that their sum of squares is closer to 1 (i.e. values that preserve the vector length better), but that sounds like a non-trivial thing to do.

astronouth7303:
We could also special case multiples of 90deg? I suspect that'll come up with more problems

nbraud:
@astronouth7303 Do you mean that special-casing multiples of 90 degrees will come with more problems, or that solving the general problem will?

nbraud:
@astronouth7303 Fixed in #89

nbraud:
PS: Rebased to solve the merge conflict with #88

nbraud:
Nevermind, I had to rebase anyway now that #79 was merged, so I could add documentation for vector-scalar division.

astronouth7303:
How likely is the doctest stuff to break? I'm nervous that Floating Point will happen.

nbraud:
@astronouth7303 Pretty unlikely?  Floating point arithmetic is still deterministic, so `doctest` won't randomly fail.

A code change can still change the precision at which some values are computed, but the examples use exact values, except for `normalize`, `scale`, and `truncate`; I guess I can fix that  :)

PR #91, opened 2018-12-14
tests: Complete the method lists

nbraud:
- [x] Add methods that do not return a `Vector2`
- [x] Use `Vector2.convert` directly, rather than a lambda, so it has a human-meaningful name.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -17,3 +17,26 @@ def units(draw, elements=st.floats(min_value=0, max_value=360)):
 def angle_isclose(x, y, epsilon = 6.5e-5):
     d = (x - y) % 360
     return (d < epsilon) or (d > 360 - epsilon)
+
+
+# List of operations that (Vector2, Vector2) -> Vector2
+BINARY_OPS = [
+    Vector2.__add__,
+    Vector2.__sub__,
+    Vector2.reflect,
+]
+
+# List of operations that (Vector2, Real) -> Vector2
+SCALAR_OPS = [
+    Vector2.scale_by,
+    Vector2.rotate,
+    Vector2.truncate,
+    Vector2.scale_to,
+]
+
+# List of operations that (Vector2) -> Vector2
+UNARY_OPS = [
+    lambda v: type(v).convert(v),
+    Vector2.__neg__,
+    Vector2.normalize,
+]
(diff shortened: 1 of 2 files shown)


tests/utils.py
```diff
@@ -36,7 +36,7 @@ def angle_isclose(x, y, epsilon = 6.5e-5):
 
 # List of operations that (Vector2) -> Vector2
 UNARY_OPS = [
-    lambda v: type(v).convert(v),
```
astronouth7303: Shouldn't this cause tests to fail?

```
class V(Vector2): pass

assert isinstance(Vector2.convert((1,2), Vector2)
assert isinstance(V.convert((1,2), V)
```

Oh, I see, it's because it's passing in a `Vector2`, so it just short-circuits and passes the instance right back out.

This is still probably wrong.

nbraud: > This is still probably wrong.
How so?  As the commit message details, tests never redefine `Vector2.convert` in a subclass, so it's equivalent, and we get `Vector2.convert` as the function name instead of `<lambda somehorriblestuff>`, in tests and benchmark output.

nbraud:
PS: The CI failure was caused by [a mypy issue](https://github.com/python/mypy/issues/220), so I added a `type: ignore` annotation as a workaround.

tests/utils.py
astronouth7303: A subclass doesn't have to override it for it to return a subclass.

If `.convert()` is called on a subclass, it'll return that subclass if construction was involved. eg, `MyVector.convert((1,2))` will return an instance of `MyVector` with no overrides necessary.

nbraud: Yes, this is true, but it's also not tested by anything in our testsuite (as far as I can tell), nor is it related to any of the tests that use `UNARY_OPS`.

Case in point, calling that lambda on something that isn't a (subclass of) `Vector2`, like a `tuple`, would expand to `tuple.convert((1, 2))` and fail:
```
>>> l = lambda v: type(v).convert(v)
>>> l((1, 0))
AttributeError: type object 'tuple' has no attribute 'convert'

>>> l = Vector2.convert
>>> l((1, 0))
Vector2(1.0, 0.0)
```

As such, I don't think there is a way for a test that uses `UNARY_OPS` (whether with the lambda or with `Vector2.convert`) to be testing that behaviour, and there definitely is no test in the testsuite that does (because it would have broken with the lambda).

Issue #90, opened 2018-12-12
Tracking test coverage and performance


nbraud:
I know we kinda-discussed that in #59, but I thought it would be useful to resume that discussion & track it in a separate issue.  (And if you feel it's inappropriate for me to bring it up again, please let me know and close the issue  :3)

I think it would be pretty nice to have coverage and performance tracking, if only because we could answer questions like “how bad is the slowdown of #89” or “is this adequately tested” without having to reinvent a new way to get that data.

I totally agree with @pathunstrom that we should minimise the amount of tooling a user has to interact with, so it should happen automatically for them.  I'd like to suggest doing it during CI, and automatically posting a message to the PR (if appropriate) with:
- a link to the full report;
- if coverage changed significantly, say it did (and by how much), congratulate the contributor on a positive change;
- same for performance.

I would happily do the tooling & integration work, if there's consensus on it being desirable (and how it should behave).  :)

astronouth7303:
I know there's services that do the coverage stuff. IDK if there exists any for performance.

nbraud:
@astronouth7303 I only know [codecov](https://codecov.io) and [coveralls](https://coveralls.io).
I seem to recall both being somewhat annoying to integrate, but at least codecov supports a wide range of CI systems (coveralls would lock us into Travis...)

I don't know of any (free for open-source) performance-tracking software available as a service. There are self-hostable options, such as [Google's Dana](https://github.com/google/dana) (which seems unmaintained since almost a year) or [codespeed](https://github.com/tobami/codespeed), but I assumed it would be preferable not to host services for this.

That's why I was suggesting making something which we can run directly in CI, and would upload a report (a human-readable one, and a machine-readable one) directly.  `perf` already supports producing machine-readable outputs, and comparing 2 reports:
```
$ # Optional: apply system settings for more stable measurements
$ sudo --preserve-env=PATH,VIRTUAL_ENV python3 -m perf system tune

$ # in branch `benchmark`
$ ./tests/benchmark.py -o benchmark_master.json --rigorous
[...]
540.54s user 27.65s system 99% cpu 9:31.29 total

$ # in the target branch; first, setup a temporary branch and merge the benchmark (only necessary because the benchmark isn't merged yet)
$ git checkout de80e8d6eb5fc109264f33522fb042a487eaa8eb -b tmp_benchmark && git merge --no-edit benchmark
$ ./tests/benchmark.py -o benchmark_unstable_rotation.json --rigorous
[...]
[...]

PR #92, opened 2018-12-14
Add a perf-based benchmark

nbraud:
First step towards #90.  Depends on #91

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 __pycache__/
 .hypothesis/
 /.mypy_cache/
+/tests/*.json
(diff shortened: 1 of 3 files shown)


nbraud:
Is there anything that needs to be done for this PR to be mergeable?

astronouth7303:
I guess my only question is, how should this be used?

nbraud:
@astronouth7303 I don't think this needs to be a full solution to #90 ?
Just having the ability to [benchmark and compare] is pretty useful, since it means we can easily check what's the performance impact of any given change (rather than me inventing a microbenchmark).

Moreover, the maintenance cost is negligible, since it's a script with less than 20 lines, which nothing depends on, and which doesn't need updating when adding new functions in `Vector2` (thanks to `tests.utils.*_OPS`).

[benchmark and compare]: https://github.com/ppb/ppb-vector/issues/90#issuecomment-449000799

PR #93, opened 2018-12-14
tests: Test that 2-ary methods work with vector-likes

nbraud:
Depends on #91

### DIFF
--- a/tests/test_convert.py
+++ b/tests/test_convert.py
@@ -1,0 +1,11 @@
+import pytest  # type: ignore
+
+from ppb_vector import Vector2
+from utils import *
+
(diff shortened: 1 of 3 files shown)


nbraud:
Is there a blocker on this?  #91 was merged.

nbraud:
PS: I just added a test for the property you mentioned in #91

astronouth7303:
At this point, the only reason I'm not merging it is #85.

nbraud:
OK, so what's holding up #85?

nbraud:
PS: That shouldn't conflict with #85, since it's only adding `test_binop_vectorlike`.
(Github displays a large diff because it doesn't detect that commits that belonged to #91 got merged.)

astronouth7303:
Oh, if it doesn't conflict, then ok.

PR #95, opened 2018-12-21
Property-based testing for truncate

nbraud:
While working on #85, I noticed that `test_scale_to_equivalent_to_truncate` was broken (it had `min_value == max_value` for the length argument), so I had to go and fix it; while I was at it, I added some more tests.

There are a few behaviours that currently make the tests fail, that should be clarified:
- [x] `test_truncate_equivalent_to_scale` detects that `x.truncate(0)` always succeeds, whereas `x.scale_to(0)` fails when `x` is null.
  Should we make scaling the null vector to length 0 succeed, or should `truncate` raise an exception?
- [x] `test_truncate_length` detects that sometimes `scale_to` rounds up, resulting in a vector slightly-longer than the desired length.
  Should we make the test more lax (for instance, checking `length <= (1 + 1e-9) * max_length`) or should `truncate` make really sure that the resulting vector isn't larger than `max_length`?

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -286,6 +286,9 @@ class Vector2:
         if length < 0:
             raise ValueError("Vector2.scale_to takes non-negative lengths.")
 
+        if length == 0:
+            return type(self)(0, 0)
+
         return (length * self) / self.length
 
     scale = scale_to
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -274,9 +274,10 @@ class Vector2:
         return self.scale(1)
 
     def truncate(self: VectorOrSub, max_length: Realish) -> VectorOrSub:
-        if self.length > max_length:
-            return self.scale_to(max_length)
-        return self
+        if self.length <= max_length:
+            return self
+
+        return self.scale_to(max_length)
 
     def scale_to(self: VectorOrSub, length: Realish) -> VectorOrSub:
         """
--- a/tests/test_vector2_truncate.py
+++ b/tests/test_vector2_truncate.py
@@ -19,7 +19,7 @@ def test_truncate_invariant(x: Vector2, max_length: float):
 
 @given(x=vectors(), max_length=floats())
 @example( # Large example where x.length == max_length but 1 * x != x
-    x=Vector2(0.0, 7.387424005855793e+62), max_length=7.387424005855793e+62
+    x=Vector2(0.0, 7.1e+62), max_length=7.1e+62
 )
 def test_truncate_equivalent_to_scale(x: Vector2, max_length: float):
     """Vector2.scale_to and truncate are equivalent when max_length <= x.length"""
--- a/tests/test_vector2_truncate.py
+++ b/tests/test_vector2_truncate.py
@@ -1,5 +1,5 @@
 import pytest  # type: ignore
-from hypothesis import assume, example, given, note
+from hypothesis import assume, event, example, given, note
 from typing import Type, Union
 from utils import floats, lengths, vectors
 
@@ -39,6 +39,7 @@ def test_truncate_equivalent_to_scale(x: Vector2, max_length: float):
     try:
         scale = x.scale_to(max_length)
     except Exception as e:
+        event(f"Exception {type(e).__name__} thrown")
         scale = type(e)
 
     if isinstance(scale, Vector2) and x.length == max_length:
--- a/tests/test_vector2_truncate.py
+++ b/tests/test_vector2_truncate.py
@@ -1,4 +1,8 @@
 import pytest  # type: ignore
+from hypothesis import assume, given
+from hypothesis.strategies import floats
+from typing import Type, Union
+from utils import vectors
 
 from ppb_vector import Vector2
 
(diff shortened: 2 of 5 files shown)


tests/test_vector2_truncate.py
```diff
@@ -10,36 +13,49 @@ def test_truncate():
...
+
+@given(x=vectors(max_magnitude=1e75), max_length=floats(min_value=0, max_value=1e75))
+def test_truncate_equivalent_to_scale(x: Vector2, max_length: float):
+    """Vector2.scale_to and truncate are equivalent when max_length <= x.length"""
+    assume(max_length <= x.length)
+
+    try:
+        truncate = x.truncate(max_length)
+    except Exception as e:
+        truncate = type(e)
```
nbraud: Is there a more pythonic way to write this?

astronouth7303: Seems right. You're testing that they either return the same value or raise the same exception.

nbraud: Yes, I think it does what I want it to, but it's fairly un-pretty

nbraud:
So, what do you think should be the correct behaviours for `truncate` and `scale_to`?  (see first message in the PR)

nbraud:
I pushed 2 commits (51bcb2b and b5c4447) with a proposed fix:
- modify `Vector2.scale_to` so it always succeeds when `length == 0` (even if `self == (0, 0)`);
- admit and document the rounding behaviour of `scale_to` and `truncate`.

Hypothesis still found an [exciting case](https://travis-ci.org/ppb/ppb-vector/jobs/472330800#L544) where `self.length == max_length` and `truncate` and `scale_to` disagree.  Will fix but my train is arriving.

nbraud:
The current CI failure is due to a bug in `Vector2.isclose`, exposed and fixed in #96

tests/test_vector2_truncate.py
```diff
@@ -19,6 +19,9 @@ def test_truncate_invariant(x: Vector2, max_length: float):
 
 
 @given(x=vectors(max_magnitude=1e75), max_length=floats(min_value=0, max_value=1e75))
+@example( # Large example where x.length == max_length but 1 * x != x
+    x=Vector2(0.0, 7.387424005855793e+62), max_length=7.387424005855793e+62
```
astronouth7303: Can you manually simplify this? Or is this the simplest form?

nbraud: I don't know if it's the simplest form, but I was unable to reproduce the issue with `7e+62` or `8e+62`

nbraud: Nevermind, the issue can be reproduced with `7.1e+62`.

tests/test_vector2_truncate.py
```diff
@@ -12,12 +12,14 @@ def test_truncate_length(x: Vector2, max_length: float):
 
 
 @given(x=vectors(), max_length=lengths())
+@settings(suppress_health_check=[HealthCheck.filter_too_much])
```
astronouth7303: Is there a way to better handle generation rather than filtering so much?

nbraud: The main issue is that I don't think there's a reasonable way to generate `x` and `max_length` in such a way that `x.length <= max_length` without introducing a bunch of bias in the test.
I guess we could increase the range on `max_length` (or `x`, for the other test), which would reduce the likelihood of the assumption not being satisfied.

In any case, Hypothesis is able to generate enough samples, the issue is only that sometimes the heath check fails, resulting in a CI failure.

nbraud: PS: I tweaked the ranged to make constraint satisfaction more likely, going from ~49% to ~100% for `test_truncate_invariant`.

Before:
```
$ pytest --hypothesis-show-statistics --hypothesis-profile ci tests/test_vector2_truncate.py -v
[...]
tests/test_vector2_truncate.py::test_truncate_invariant:

  - 1000 passing examples, 0 failing examples, 1034 invalid examples
  - Typical runtimes: < 1ms
  - Fraction of time spent in data generation: ~ 60%
  - Stopped because settings.max_examples=1000
```

After:
```
$ pytest --hypothesis-show-statistics --hypothesis-profile ci tests/test_vector2_truncate.py -v
[...]
tests/test_vector2_truncate.py::test_truncate_invariant:

  - 1000 passing examples, 0 failing examples, 3 invalid examples
  - Typical runtimes: < 1ms
  - Fraction of time spent in data generation: ~ 61%
  - Stopped because settings.max_examples=1000

tests/test_vector2_truncate.py::test_truncate_equivalent_to_scale:

  - 1000 passing examples, 0 failing examples, 62 invalid examples
  - Typical runtimes: < 1ms
  - Fraction of time spent in data generation: ~ 58%
  - Stopped because settings.max_examples=1000
```

nbraud:
PS: This got a bit out of hand (17 commits, not counting the merge from `master`), sorry about that.
I merged from `master` rather than rebasing, because the conflict resolution was a bit annoying and I couldn't justify to myself the extra effort in that case.  :3

ppb_vector/vector2.py
```diff
@@ -288,6 +289,9 @@ def scale_to(self: VectorOrSub, length: Realish) -> VectorOrSub:
         if length < 0:
             raise ValueError("Vector2.scale_to takes non-negative lengths.")
 
+        if length == 0:
+            return type(self)(0, 0)
```
nbraud: I just noticed that the testsuite doesn't fail if we return `Vector2(0, 0)` there.

That's probably an opportunity to make more tests in `test_typing` use Hypothesis, so we can hit all codepaths in the methods under test, not the single one that's exercised by the hardcoded inputs, but that feels out-of-scope here, and like something that should be discussed (esp. since I have very little understanding of Python's object model).

@astronouth7303 @pathunstrom Please open a new issue for this if you think it's worth fixing.

nbraud:
I just realised there was no reply to [last week's message](https://github.com/ppb/ppb-vector/pull/95#issuecomment-449963004).

Are the proposed behaviours the one we want?

ppb_vector/vector2.py
astronouth7303: I'm ok with leaving the type-oriented stuff largely manual for now, since there's a relatively small number of possible types accepted, unless you think there's significant value in doing a hypothesis thing with it.

astronouth7303:
They seem like reasonable behaviors.

ppb_vector/vector2.py
nbraud: See proposal in #101 

Given that moving those tests to Hypothesis didn't bring in extra complexity, and that it makes them much more useful (since we are now testing all code paths, instead of a fixed one), I think it would be reasonable to have.

nbraud:
OK, what's the blocker here?  CI is green, and you agreed the implemented behaviours are correct.

PR #97, opened 2018-12-29
Use consistent strategies for tests

nbraud:
- [x] `utils.vectors`: Set the default `max_magnitude` to 10⁷⁵.
  No test uses a smaller value, and larger vectors are meaningless: the observable Universe is 9×10²⁶ meters across...
- [x] Introduce convenience strategies for `angles` and `lengths`.
  This should make it more obvious which inputs are what, and that should help with #85.
- [x] Simplify `utils.units`.

### DIFF
--- a/tests/test_vector2_angle.py
+++ b/tests/test_vector2_angle.py
@@ -47,7 +47,7 @@ def test_angle_additive(left, middle, right):
     assert angle_isclose(lm + mr, lr)
 
 @given(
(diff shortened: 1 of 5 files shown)


nbraud:
Actually, I merged all the in-flight pull requests in a temporary branch, and it wasn't too bad.
Still, it would be nice if we could move a bit towards closing/merging.

PR #99, opened 2019-01-02
Add negative tests for Vector2.isclose

nbraud:
- [x] Refactor after #97 merge.
- [x] Negative test for `abs_tol`
- [x] Negative test for `rel_tol`
- [x] Fix [flakyness](https://travis-ci.org/ppb/ppb-vector/jobs/474457841#L699) in `test_isclose_abs_error`


nbraud:
The negative example for `rel_tol` is [heavily commented](https://github.com/ppb/ppb-vector/pull/99/files#diff-d1af895e065c7292fd40333ae97cc313R56) because it was somewhat difficult to figure out.

nbraud:
I fixed the issue that broke #101, so this should be good now.

PR #101, opened 2019-01-02
tests/typing: Replace hardcoded inputs with Hypothesis parameters

nbraud:
The testsuite now catches the [potential bug](https://github.com/ppb/ppb-vector/pull/95#discussion_r244650431) discussed in #95

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -59,7 +59,20 @@ class Vector2:
     x: float
     y: float
 
-    def __init__(self, x: typing.SupportsFloat, y: typing.SupportsFloat):
+    @typing.overload
+    def __init__(self, x: typing.SupportsFloat, y: typing.SupportsFloat): pass
+
+    @typing.overload
+    def __init__(self, other: VectorLike): pass
+
+    def __init__(self, *args):
+        if len(args) == 1:
+            x, y = Vector2.convert(args[0])
+        elif len(args) == 2:
+            x, y = args
+        else:
+            raise TypeError(f"Expected 1 vector-like or 2 float-like arguments, got {len(args)}")
+
         try:
             # The @dataclass decorator made the class frozen, so we need to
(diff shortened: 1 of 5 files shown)


nbraud:
@astronouth7303 Following our conversation last night, I tried to use Hypothesis in `test_typing.py` and it turned out not to make it more complicated; in fact, I think the type annotations are a readability improvement.
What do you think?

nbraud:
The CI failure is unrelated, and I added it in #99 as an explicit example so we can fix it there.

nbraud:
There's a fix in #99 now

astronouth7303:
Naughty! Naughty! There's changes to the signature of `__init__()` mixed in this testing PR! :stuck_out_tongue: However, this is more Pythonic.

The tests themselves seem solid.

nbraud:
@astronouth7303 Yeah, it seemed more reasonable to support building a vector from a vector-like than hack around the lack of support, though I just realised I could have used `st.builds(V1.convert, _)`.

Anyhow, I assumed you would look at the commits in the PR, and it's in a separate commit so it can easily be pulled out if you'd rather have it be a separate discussion.

nbraud:
PS: Now that the fix for the CI failure is merged, can you re-run the build/tests?

nbraud:
OK, #102 should finally(!) squash that CI failure.

nbraud:
CI is now green.

nbraud:
Urg, I forgot that Github gets confused when one pushes a merge commit to a PR branch; going to rebase and force-push...

nbraud:
Since this wasn't merged yet, I included some conversion test cleanups.
The tests caught that `dict(some_vector2)` didn't work, so I also implemented the `mapping` protocol.

nbraud:
Rebased because of the last merges, added one more test (see last commit: 
2f2bfdb)

PR #102, opened 2019-01-03
Make the tests for isclose less flaky

nbraud:
I was able to run Hypothesis, in the CI profile, ~200~ 500 times in a row without encountering a single failure:
```
$ for i in {1..200}; pytest -v --hypothesis-profile ci tests/test_vector2_isclose.py || { echo $i ; break }
```


nbraud:
Update: I was able to run Hypothesis (still in CI profile) 300 more times, so half-a-million valid inputs were generated for each test without encountering failures anymore.

PR #103, opened 2019-01-05
Add tests for negation

nbraud:
- [x] Replace the parameterised negation test with an Hypothesis one.
- [x] Generalise it to scalar multiplication, obsoleting another parameterised test.
- [x] Add a couple more properties to test about negation (involutive, additive inverse)


nbraud:
I greatly enjoyed testing more things while only adding a single line  ;)

PR #104, opened 2019-01-05
Implement the mapping protocol

nbraud:
Follow-up on #101; the commits from that feature branch are included, so please merge it first.

- [x] Add tests converting vectors to and from `dict`s.
- [x] Implement the `mapping` protocol, so `dict(some_vector)` works.

The PR seems to expose a bug in `mypy`, but this is snek sourcery beyond my ken, so I'd like one of you to have a look.

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -172,6 +172,9 @@ class Vector2:
         """Perform a division between a vector and a scalar."""
         return type(self)(self.x / other, self.y / other)
 
+    def keys(self: VectorOrSub) -> typing.Sequence[str]:
+        return ['x', 'y']
+
     def __getitem__(self: VectorOrSub, item: typing.Union[str, int]) -> Realish:
         if hasattr(item, '__index__'):
             item = item.__index__()  # type: ignore
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -60,7 +60,20 @@ class Vector2:
     x: float
     y: float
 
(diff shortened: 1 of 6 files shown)


nbraud:
`mypy`'s documentation discusses [structural subtyping and protocols](https://mypy.readthedocs.io/en/latest/protocols.html) but there's no discussion of the mapping protocol.

astronouth7303:
Is there a rationale for this, besides "to enable bidirectional conversion between dicts and vectors"?

astronouth7303:
The basic problem with this is that `Vector2` already implements the Sequence protocol.

The Sequence protocol is incompatible with the Mapping protocol because they have different ideas of `__iter__()`. (Sequence would generate the contents, Mapping would generate the keys.)

The error appears to stem from a limited understanding of what `dict()` will accept on the part of mypy. Specifically, it should accept:

1. `Mapping[Any, Any]`
2. `Iterable[Tuple[Any, Any]]`
3. Generic keyword arguments

That, or because you don't inherit from `collections.abc.Mapping`, mypy doesn't check form 1. But I would caution against this, see the Sequence/Mapping conflict above.

Given that I'm not sure it's possible for a single object to correctly implement both Sequence and Mapping at the same time, I think I'm going to have to reject this PR, unless you can come up with a way to rectify this apparent conflict.

nbraud:
@astronouth7303 Yes, this was indeed to enable bidir conversions, since we already provide the others.
I think I understand the issue, thanks for the explanation  <3

PR #105, opened 2019-01-06
Travis: Run tests under PyPy

nbraud:
- [x] Extract the environment setup script to a separate file.
- [x] Improve the output from the setup script.
- [x] Support installing PyPy.
  Thanks to `matrix.fast_finish`, the Github check will go green without waiting on the PyPy builds.
- [x] Move downloads outside of the source directory.
   The pollution of the source dir caused pytest to scan the whole PyPy source tree.
- [x] Add the `sudo: no` setting to run on Travis' container infrastructure.
  We weren't doing anything as root, and container builds start much faster.

### DIFF
--- a/.ci/install-python.sh
+++ b/.ci/install-python.sh
@@ -9,13 +9,31 @@ if [[ "$TRAVIS_OS_NAME" != linux ]]; then
    exit 1
 fi
 
-MINICONDA_OS=Linux
-URL="https://repo.continuum.io/miniconda/Miniconda3-latest-${MINICONDA_OS}-x86_64.sh"
-wget "${URL}" -O miniconda.sh
-bash miniconda.sh -b -p "$HOME/miniconda"
-export PATH="$HOME/miniconda/bin:$PATH"
-conda config --set always_yes yes --set changeps1 no
-conda update -q conda
-conda info -a
-conda create -q -n test-environment "python=$PYTHON"
-source activate test-environment
+if [[ "$PYTHON" =~ pypy-* ]]; then
+  # Download & install a prebuilt pypy snapshot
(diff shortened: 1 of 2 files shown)


nbraud:
I mostly did this because I wanted to see if the `py3.6` branch PyPy can run `ppb_vector`, and it does!

It's also interesting to see that it gets a 40% speedup on `pytest`, and probably more if we increase `max_examples` in the Hypothesis profile, so that might be something to keep in mind once PyPy ships a stable version compatible with Py3.6

nbraud:
PS: Despite the larger number of builds being run, the wall clock time for the whole CI run was reduced by 22%.

astronouth7303:
What's up with pypy-nightly?

nbraud:
@astronouth7303 The `trunk` build doesn't seem to include the `ensurepip` module.  I asked some PyPy folks whether that was on purpose, or a bug in the nightly build.

astronouth7303:
Ok. I know we don't want to block on it, but we should keep an eye on it, and at least start it as green.

astronouth7303:
No, I'd rather keep it. Especially if trunk is going to be superceded by 3.6.

nbraud:
@astronouth7303 Turns out [`trunk` is PyPy 2](https://pypy.org/download.html#building-from-source), so that indeed won't work.
Looks like `py3.6` is the only PyPy branch we want currently, so I will drop `trunk` and that should be mergeable.

nbraud:
@astronouth7303 OK, I removed the `pypy-trunk` build, removed the PyPy-specific Hypothesis profile (that made it generate 10× more examples under PyPy), and rebased to clean up the history.
This should be good to go.

PR #106, opened 2019-01-11
Use the dataclass helper and make vectors immutable

nbraud:
Following a discussion with @astronouth7303.

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -1,6 +1,7 @@
 import typing
 import collections
 import functools
+import dataclasses
 from dataclasses import dataclass
 from math import acos, atan2, cos, degrees, hypot, isclose, radians, sin, copysign, sqrt
 from collections.abc import Sequence, Mapping
@@ -76,6 +77,8 @@ class Vector2:
             object.__setattr__(self, 'y', float(y))
         except ValueError:
             raise TypeError(f"{type(y).__name__} object not convertable to float")
+
+    update = dataclasses.replace
 
     @classmethod
     def convert(cls: typing.Type[VectorOrSub], value: VectorLike) -> VectorOrSub:
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -1,7 +1,7 @@
 import typing
 import collections
 import functools
-from dataclasses import dataclass
+from dataclasses import dataclass, FrozenInstanceError
 from math import acos, atan2, cos, degrees, hypot, isclose, radians, sin, copysign, sqrt
 from collections.abc import Sequence, Mapping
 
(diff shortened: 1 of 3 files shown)


nbraud:
Right, this fails on 3.6 because `dataclasses` was introduced in 3.7.
I guess this is a reasonable justification to reject for now, or at least delay until there are other/stronger reasons to require 3.7 or later?

nbraud:
PS: There is a version of `dataclasses` [on PyPI](https://pypi.org/project/dataclasses/), which should work on 3.6.

astronouth7303:
Add `dataclasses; python_version < '3.7'` to the requirements list.

requirements.txt
```diff
@@ -0,0 +1 @@
+dataclasses
```
astronouth7303: No. This goes in `setup.py`.

ppb_vector/vector2.py
```diff
@@ -56,35 +57,37 @@ def _find_lowest_vector(left: typing.Type, right: typing.Type) -> typing.Type:
...
+@dataclass(eq=False, frozen=True, init=False, repr=False)
 class Vector2:
     x: float
     y: float
 
     def __init__(self, x: typing.SupportsFloat, y: typing.SupportsFloat):
         try:
-            self.x = x.__float__()
-        except AttributeError:
+            object.__setattr__(self, 'x', float(x))
```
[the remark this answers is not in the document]
astronouth7303: Ok, the `dataclasses` documentation calls this out. Please add a comment to the effect of "this is the prescribed pattern for initializing frozen dataclasses [link]"

ppb_vector/vector2.py
```diff
@@ -56,35 +57,37 @@ def _find_lowest_vector(left: typing.Type, right: typing.Type) -> typing.Type:
         return _find_lowest_type(left, right)
 
 
+@dataclass(eq=False, frozen=True, init=False, repr=False)
```
astronouth7303: This should probably have `eq=True` for reasons of hashing?

nbraud: That should result in a `TypeError` since `Vector2` defines it's own `__eq__`.

Moreover, I don't think we can make it an hashable type, as AFAIK `hash(vector) == hash(vector_like)` would need to hold wherever `vector == vector_like`.

nbraud:
@astronouth7303 I didn't suggest requiring users to know about `dataclass`, but more that `update = dataclasses.replace` would be a nice convenience function to expose.

nbraud:
OK, rebased now that CI was green again and that your comments were handled (except for `update`).

ppb_vector/vector2.py
```diff
@@ -77,6 +78,8 @@ def __init__(self, x: typing.SupportsFloat, y: typing.SupportsFloat):
         except ValueError:
             raise TypeError(f"{type(y).__name__} object not convertable to float")
 
+    update = dataclasses.replace
```
astronouth7303: How does this look in `help()`/pydoc?

ppb_vector/vector2.py
astronouth7303: Yes, for this you want to key the same, two equal things must hash the same.

In this case, I don't think we want to key the same to anything not an actual vector? And this becomes a problem not because we expect users to key dictionaries on vectors, but because of things like `functools.lru_cache`, which use dictionaries under the hood.

ppb_vector/vector2.py
nbraud:
```python3
>>> from ppb_vector import Vector2
>>> help(Vector2.update)
Help on function replace in module dataclasses:

replace(obj, **changes)
    Return a new object replacing specified fields with new values.
    
    This is especially useful for frozen classes.  Example usage:
    
      @dataclass(frozen=True)
      class C:
          x: int
          y: int
    
      c = C(1, 2)
      c1 = replace(c, x=3)
      assert c1.x == 3 and c1.y == 2
```

That's indeed not very nice.  OTOH, `help()` is currently useless on `Vector2` because:
- Type constructs like `VectorLike` are expanded into an illegible mess.
- Description & documentation are in the `README` rather than in docstrings.

I think it would be worth fixing, but it's way out of scope for this PR.

astronouth7303: Fair enough. We should figure out a way to handle that.

PR #107, opened 2019-01-15
Convert equality tests to Hypothesis

nbraud:
- [x] Replace equality tests that used hardcoded values with Hypothesis versions
- [x] Remove unnecessary definition of `__ne__`

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -185,11 +185,12 @@ class Vector2:
         return f"{type(self).__name__}({self.x}, {self.y})"
 
     def __eq__(self: VectorOrSub, other: typing.Any) -> bool:
-        if is_vector_like(other):
+        try:
             other = Vector2.convert(other)
             return self.x == other.x and self.y == other.y
-        else:
-            return False
+
+        except (TypeError, ValueError):
+            return NotImplemented
 
     def __iter__(self: VectorOrSub) -> typing.Iterator[Realish]:
         yield self.x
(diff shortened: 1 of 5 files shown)


nbraud:
Found a bug in `is_vector_like` while adding a test that `x != "foo"` (when `x` is a vector).

I initially added a slow-but-correct fix (call Vector2.convert, check whether an exception is raised), confirmed it fixed the bug, then refactored the 2 call sites not to need `is_vector_like` and deleted that function.

ppb_vector/vector2.py
```diff
@@ -183,14 +177,12 @@ def __repr__(self: VectorOrSub) -> str:
         return f"{type(self).__name__}({self.x}, {self.y})"
 
     def __eq__(self: VectorOrSub, other: typing.Any) -> bool:
-        if is_vector_like(other):
+        try:
             other = Vector2.convert(other)
             return self.x == other.x and self.y == other.y
```
astronouth7303: I would normally do this in the form:

```
try:
    other = Vector2.convert(other)
except ...:
    return NotImplemented
else:
    return self.x == other.x and self.y == other.y
```

But on reflection, in situations where the comparison raises an error, I think that returning `NotImplemented` would be the right thing. (Although also indicative of a bug in `.convert()`.)

nbraud: Mmmmh, yes, this captures my intent more accurately *and* I think what is going on is clearer.
Thanks <3

PR #108, opened 2019-01-18
tests/typing: Add tests for conversions to/from dictionary

nbraud:
Second take on #104 ; still based on #101.

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -124,6 +124,9 @@ class Vector2:
         # benefit, according to microbenchmarks.
         return hypot(self.x, self.y)
 
+    def asdict(self) -> typing.Mapping[str, float]:
+        return {'x': self.x, 'y': self.y}
+
     def __len__(self: VectorOrSub) -> int:
         return 2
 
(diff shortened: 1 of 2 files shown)


ppb_vector/vector2.py
```diff
@@ -94,6 +107,9 @@ def length(self) -> float:
         # benefit, according to microbenchmarks.
         return hypot(self.x, self.y)
 
+    def asdict(self) -> typing.Mapping[str, float]:
```
nbraud: Should this be a `property`?

nbraud:
Rebased on #101, removed a bit of fluff from the dictionary test.

ppb_vector/vector2.py
astronouth7303: Not normally.

astronouth7303:
Can you rebase again? Looks like another branch is still mixed in.

ppb_vector/vector2.py
nbraud: OK.

nbraud:
@astronouth7303 That's because Github doesn't notice you merged most of those commits as part of #101.  I can rebase indeed.

nbraud:
Done.

PR #111, opened 2019-02-04
CI: Run flake8 (plus various plugins) during tests

nbraud:
- Superset of #116 :
  - [x] Extract tests & QA tools execution to a `tests.sh` script.
  - [x] Reorder execution order to fail faster.
  - [x] Eliminate mutable default parameters (found by `flake8-bugbear`)
  - [x] Uniformize the ordeer of `import` lines.
  - [x] Uniformize the use of trailing commas

- Automatically enforce the things that were fixed in #116 :
  - [x] Check basic lints with `flake8`
  - [x] Check for PEP 8 [naming conventions](https://www.python.org/dev/peps/pep-0008/#naming-conventions)
    This caught an issue that should have been fixed in #85 (a scalar named `l`); this likely happened because the test in question was added concurrently to #85.
  - [x] Enable `flake8-bugbear` and its “opinionated” checks
    This caught a number of instances where mutable data was used in default parameters.
  - [x] Check for trailing commas where useful
  - [x] Enforce a consistent import order

[PEP 8]: https://www.python.org/dev/peps/pep-0008/
[`black`]: https://github.com/ambv/black


pathunstrom:
How do you run these tests in a Windows environment?

pathunstrom:
Additionally, I am strongly against a linter as a dev requirement.

nbraud:
@pathunstrom The commands moved from being in `.travis.yml` (which contains Bash commands) to being in a shell script; in neither case can they be readily invoked in a (native) Windows environment, so that's not a regression.

nbraud:
> I am strongly against a linter as a dev requirement.

Could you provide a rationale?

pathunstrom:
I guess last quest, you mention mutable defaults, but I don't see any such changes in the code, am I missing them?

ppb_vector/vector2.py
```diff
@@ -242,7 +238,8 @@ def angle(self: VectorOrSub, other: VectorLike) -> float:
         return rv
 
     def isclose(self: VectorOrSub, other: VectorLike, *,
-                abs_tol: Realish=1e-3, rel_tol: Realish=1e-06, rel_to: typing.Sequence[VectorLike]=[]) -> bool:
+                abs_tol: Realish = 1e-3, rel_tol: Realish = 1e-06,
+                rel_to: typing.Sequence[VectorLike] = ()) -> bool:
```
nbraud: Mutable default removed here.

tests/utils.py
```diff
@@ -1,41 +1,60 @@
...
-def isclose(x, y, abs_tol: float=1e-9, rel_tol: float=1e-9, rel_exp: float=1,
-            rel_to: Sequence[Union[float, Vector2]]=[]):
+
+def isclose(
+    x,
+    y,
+    abs_tol: float = 1e-9,
+    rel_tol: float = 1e-9,
+    rel_exp: float = 1,
+    rel_to: Sequence[Union[float, Vector2]] = (),
```
nbraud: Mutable default removed here

nbraud:
@pathunstrom I found 2 at a quick glance, can't recall offhand if there were more.
You can check commit 36976d9, which introduced that lint check (it's mentioned in the commit message).

tests/utils.py
pathunstrom: Thanks for pointing them out.

astronouth7303:
Based on discord discussion, let's break this up.

I'm willing to accept all of the source changes made by the formatting in a PR. I don't disagree with any of it.

However, we need to have further discussions on the tooling changes. I think that `test.sh` is good (although it needs a matching .bat), I think we agree that some level of lint is good, and that having that lint as part of the local development process as simply as possible is good.

What needs further discussion is what the exact lint ruleset should be.

nbraud:
> [`test.sh`] needs a matching `.bat`
Can this either happen in a follow-up PR or can one of you commit it to this branch?
There's no Windows-based CI here, and I don't have any Windows system to test with.

nbraud:
> What needs further discussion is what the exact lint ruleset should be.

I think we ought to iterate as false-positives come up?
Every single `flake8` plugin that was added here found issues, so they are arguably useful, and I find it hard to reason about single lint rules in the abstract (though you two probably have more experience with Python linters)

nbraud:
I just rebased so CI would rerun with the fix from #112

nbraud:
The latest commit (40a0269) fixed the PyPy build: `test.sh` was accidentally attempting to run `mypy` under `pypy3`.

nbraud:
> What needs further discussion is what the exact lint ruleset should be.

There's been no discussion, and no response to my comments.

nbraud:
The CI issue seems to be the same issue #112 should have fixed  :(

nbraud:
Confirmed: we have [`attrs 17.3.0`](https://travis-ci.org/ppb/ppb-vector/jobs/497736969#L557) installed along [`pytest 3.10.1`](https://travis-ci.org/ppb/ppb-vector/jobs/497736969#L582) even though `pytest` requires `attrs >= 17.4.0` since [version 3.5.0](https://github.com/pytest-dev/pytest/pull/3228).

nbraud:
OK, looks like #114 fixed the CI issue (hopefully for good...)

nbraud:
Ping?

nbraud:
@astronouth7303 @pathunstrom I did what I suggested and split off the actual changes into #116.
This branch contains #116, plus the commits that enforce those lints in CI.

nbraud:
OK, rebased because Github was still displaying the changes from #116

nbraud:
Rebased to solve the merge conflict, and fixed the lints that broke (because I can't write correctly-styled Python unassisted).

nbraud:
So, about that discussion on which lint rules are desirable, you said it's needed but all we got is me monologuing for a month-and-a-half.

astronouth7303:
Looks like hypothesis might have found a bug? Care to investigate?

nbraud:
@astronouth7303 Was the failure you mentioned one of the things that we fixed recently?

astronouth7303:
I can't find the CI that's in reference to, so I have no idea.

nbraud:
Rebased again, to deal with merge conflicts.

astronouth7303:
Just going for it.

bors r+

nbraud:
@astronouth7303 I fixed the lint issues that got introduced into `master` since the last rebase.

ppb_vector/vector2.py
```diff
@@ -487,7 +487,7 @@ def truncate(self: Vector, max_length: typing.SupportsFloat) -> Vector:
 
         Note that :py:meth:`vector.scale(max_length) <scale>` is equivalent to
         :py:meth:`vector.truncate(max_length) <truncate>` when
-        :py:meth:`max_length \< vector.length <length>`.
+        :py:meth:`max_length < vector.length <length>`.
```
astronouth7303: This doesn't render right in Sphinx.

nbraud: Ohno.  I thought I would dodge the issue by using a unicode `<` sign, rather than the angled braket  ;_;
Fixed with a different symbol  O:3

astronouth7303:
NERD.

bors r+

nbraud:
@astronouth7303 Guilty as charged <3

nbraud:
On one hand, ≨ is a bit of an arcane rune, OTOH I like that it's extra explicit about `v.truncate(l)` and `v.scale_to(l)` not being always equivalent when `l == v.length` (Hypothesis found test vectors such that `v.x != 1.0 * v.x` :sob: )

PR #116, opened 2019-03-15
Support running tests locally, fix various warts

nbraud:
This is a subset of #111 

- [x] Extract tests & QA tools execution to a `tests.sh` script.
- [x] Reorder execution order to fail faster.
- [x] Eliminate mutable default parameters (found by `flake8-bugbear`)
- [x] Uniformize the ordeer of `import` lines.
- [x] Uniformize the use of trailing commas

The enforcement of those lints is left up to #111, as the discussion on which lint rules to include or not failed to happen.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ cache:
 
 before_install:
   - source ./.ci/install-python.sh
+  - export TRAVIS_OS_NAME PYTHON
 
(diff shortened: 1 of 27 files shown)


nbraud:
@astronouth7303 I think you accidentally closed without merging.

PR #118, opened 2019-03-16
Get rid of Realish

nbraud:
- [x] Use the exact return type `float` rather than `Realish`.
  No change required except in type annotations.
- [x] In argument types, use `typing.SupportsFloat` rather than `Realish`.
  Also added the necessary conversions to `float`.
- [x] Remove the unused `Realish` definition.

This is entirely backwards compatible:
- `Realish` is a subtype of `typing.SupportsFloat`, so any caller that provided a `Realish` will be fine.
- `float` is a subtype of `Realish`, so any caller that expected a `Realish` return value will be fine.

As a bonus, this makes the type annotations in pydoc nicer: pydoc was inlining the definition of `Realish`, so there were `Union[int, float]` everywhere.

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -145,7 +145,7 @@ class Vector2:
             return NotImplemented
         return rtype(self.x - other.x, self.y - other.y)
 
-    def dot(self: VectorOrSub, other: VectorLike) -> Realish:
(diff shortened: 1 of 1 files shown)


nbraud:
@astronouth7303 You merged `master` into this; did you mean to merge it into `master` instead?

astronouth7303:
Oops, I did the online merge conflict thing.

PR #120, opened 2019-03-16
tests/rotate: Test trigonometric invariance (+360° changes nothing)

nbraud:
This is probably somewhat-overkill, but I realised it was something important to check, that was missing

PR #121, opened 2019-03-21
Use Cirrus CI

### LINKED ISSUE #117: Move CI away from Travis
Since its acquisition, Travis CI fired most of its core engineering team, and the free service has become unbearably slow (dozens of minutes before a job even starts).

Given the impact on development velocity and QA (a couple of PRs were just merged without waiting for CI), and given the apparent lack of future for travis-ci.org, I suggest we move somewhere else.

I seem to recall @astronouth7303 suggesting [Cirrus CI](https://cirrus-ci.org/), which has native support for Linux, macOS, Windows and FreeBSD.  As a plus, having something container-based means we can stop piling up hacks to deal with Travis' environments.

nbraud:
- Run CI on multiple operating systems
  - [ ] FreeBSD
  - [x] Linux
  - [x] macOS
  - [x] Windows
- Run on multiple Python interpreters
  - [x] CPython 3.6
  - [x] CPython 3.7
  - [ ] PyPy 3.6
- [x] Separate lints and tests
  Lints can now be executed with `lint.sh`, and their dependencies are in `lint-requirements.txt` (plus `dev-requirements.txt`).
- [x] Run `lint.sh` once, in a separate task from tests
  This avoids running them (2 Python versions)×(4 operating systems) = 8 times.

Many thanks to @duckinator, the smollest dog, whose [Cirrus config](https://github.com/duckinator/emanate/blob/master/.cirrus.yml) I copied in large parts :innocent: 

Closes #117


nbraud:
@astronouth7303 Could you turn on Cirrus CI for the repo?

nbraud:
FreeBSD support will need me to refactor `test.sh` to support calling Python with a custom interpreter name (like `python36` or `python37`).

nbraud:
> Python interpreter `python` not found.
Grrrrr Travis  :(

pathunstrom:
Yikes. Thanks for doing this effort!

astronouth7303:
Should be hooked up now. Might need a dummy commit to enable it.

And yeah, configuring CIs is often annoying.

nbraud:
@pathunstrom I straight-up lifted the gnarliest bits from @duckinator's Cirrus config, so it wasn't as bad as it could have been.

I actually had a stab at making a Cirrus config for testing Python things in a dummy repo, but I ended up not having the spoons to make it work, so I'm happy I could just use that as a starting point.

nbraud:
Wheee, the build isn't entirely red!  :tada:ish

nbraud:
```
$ pip install -e .
Obtaining file:///tmp/cirrus-ci-build
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/cirrus-ci-build/setup.py", line 21, in <module>
        long_description=readme(),
      File "/tmp/cirrus-ci-build/setup.py", line 6, in readme
        return file.read()
      File "/usr/local/lib/python3.6/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 3527: ordinal not in range(128)
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/cirrus-ci-build/
```

Could it be that Python depends on the current locale to parse source code?  :scream_cat:

astronouth7303:
```
 File "/tmp/cirrus-ci-build/setup.py", line 6, in readme
        return file.read()
```

No, just uses the system encoding (current locale) for opening files in text mode.

.cirrus.yml
```diff
@@ -1,6 +1,9 @@
 lint_task:
   container:
     image: python:3.7-slim
+  env:
+    LC_ALL: C.UTF-8
```
astronouth7303: I would have actually fixed this in `setup.py`

nbraud: How?

nbraud:
OK, we have minimal functionality (Linux lints + test) working, and working quite fast.
Let's see how long the rest takes.

.cirrus.yml
astronouth7303: `open(<file>, encoding='utf-8')`

For more details, see https://docs.python.org/3/library/functions.html#open

.cirrus.yml
```diff
@@ -0,0 +1,106 @@
...
+    - pip install -e .
+
+  script:
+    - python3 --version
+    - pip list
+    - ./test.sh
+
+Windows_task:
+  windows_container:
+    image: cirrusci/windowsservercore:2019
```
astronouth7303: There are [Python+Windows containers](https://hub.docker.com/_/python/).

.cirrus.yml
```diff
@@ -15,6 +15,7 @@ lint_task:
 
 
 FreeBSD_task:
+  allow_failures: $PYTHON == 3.7
```
astronouth7303: Can I yell at FreeBSD about this? They don't have pip3.7 packages yet and 3.8 is around the corner?

nbraud: @astronouth7303 IIRC, it's just that there was some very recent shuffling around in how Python versions are handled in the ports tree (switching to using flavors of a single source port), which involves rebuilding all those packages, and most py37 packages haven't been rebuilt yet:
```
% pkg search ^py37 | wc -l
       6
% pkg search ^py36 | wc -l  
    2038
```

astronouth7303: Ahhh, so is there an ETA on those packages getting built?

nbraud: No idea offhand

nbraud:
I seem to be having trouble with the FreeBSD build (though I suspect it's trouble correctly activating the venv in Cirrus' scripting environments).
Should I postpose FreeBSD builds to a later PR?

astronouth7303:
I'm pretty ok leaving freebsd off for now.

It looks like the Mac failure is a legit failure. What's up with that?

nbraud:
@astronouth7303 I suspect the test I added in #120 must be slightly flaky :crying_cat_face: 
I guess I get to fix it (again!)  :O

nbraud:
OK, I won't insist on adding tests under PyPy right now, but I pushed that off to a separate PR.  :)

nbraud:
This should be good to go once #123 lands (it fixes the flaky tests).
I pushed off testing under FreeBSD to #122, and under PyPy to #124

nbraud:
@astronouth7303 The failure on Travis is due to it finding *another* test issue, but this time in `test_rotation_stability` (which has been in there for a looong while)...

I don't think I'm dealing with this specific issue tonight, so proceed as you deem fit:
- either let this wait until the unrelated testsuite bug is fixed,
- ignore the failure in Travis and merge, or
- fix it yourself and then merge (after confirming CI is fully green).

astronouth7303:
I'm going to merge this on the assumption that because this should contain no substantive changes to travis, this is an intermittent error.

However, you're still the numerical wizard and you still need to fix it.

nbraud:
@astronouth7303 I'm not a numerical wizzard, I'm a mathemagician  :D

nbraud:
@astronouth7303 I guess you can force a re-run of the Travis build, to check it is indeed a transient issue.

nbraud:
Ping?

astronouth7303:
Should be good now, although I would prefer one more merge from master? I like seeing things green before I hit the merge button.

nbraud:
@astronouth7303 I just did it, but again you could also have done it yourself (maintainers have write access to PR branches by default)

nbraud:
I'm amused that Cirrus finished all Linux tests before Travis finished a single one.

PR #124, opened 2019-03-21
Cirrus CI: Run tests under PyPy 3.6

nbraud:
Follow-up on #121

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -43,13 +43,15 @@ Linux_task:
       - image: python:3.6-slim
       - image: python:3.7-slim
       - image: python:3.8-rc-slim
+      - image: pypy:3.6-slim
 
   install_script:
     - pip install --upgrade-strategy eager -U -r dev-requirements.txt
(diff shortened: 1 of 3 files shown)


nbraud:
Oops, I didn't notice because it was marked as an allowed failure in Travis (which doesn't report status on individual jobs), but I think `tests/memory` is broken under PyPy (because `pympler` doesn't work there)  :(

astronouth7303:
Indeed it doesn't.

I'm pretty ok with not running that test under pypy? It's a test for performance, not correctness, so I consider it significantly less important. That is, I care much more that it works under PyPy than I do that it has the same performance characteristics.

nbraud:
Sure, I think skipping that test under PyPy (or at least allowing it to fail) is the right thing to do; I just left a comment so I'll know about it when I pick it up again.

nbraud:
@astronouth7303 I just fixed the merge conflict (happened since #133 also touched the list of container images), this should now be good to go.  :)

PR #130, opened 2019-03-23
Add Sphincs-based documentation

### LINKED ISSUE #109: Improve docstrings & type signatures
From #106 :

> That's indeed not very nice. OTOH, help() is currently useless on Vector2 because:
>
> - Type constructs like VectorLike are expanded into an illegible mess.
> - Description & documentation are in the README rather than in docstrings.
>
> I think it would be worth fixing, but it's way out of scope for this PR.

nbraud:
Closes #109


nbraud:
@astronouth7303 Regarding what I mentioned, there's a [RTD issue] opened for a while, about adding CI-style documentation builds, and someone made [`rtd-bot`] as a stopgap – it does not (seem to) provide a Github Check, but at least it triggers a build of the documentation and links them in the PR (and it want annoying amounts of privileges, both Github-side and RTD-side).

[RTD issue]: https://github.com/rtfd/readthedocs.org/issues/1340
[`rtd-bot`]: https://github.com/KengoTODA/rtd-bot#get-url-of-staging-document-when-you-review-changes

astronouth7303:
We can probably just add a cirrus task to build docs? Given that you also need to be able to build them locally.

nbraud:
@astronouth7303 Yeah, I guess this will do as a stop-gap.

It's kinda annoying since they won't be published anywhere, so someone reviewing a PR can't go look at how a documentation change renders without checking out and building locally, but the “real” solution would be for RTD to provide a CI-like integration.

ppb_vector/vector2.py
```diff
@@ -51,6 +51,46 @@ def _find_lowest_vector(left: typing.Type, right: typing.Type) -> typing.Type:
...
+        >>> from ppb_vector import Vector2
+        >>> Vector2(3, 4)
+        Vector2(3.0, 4.0)
+
+    :py:class:`Vector2` implements many convenience features, as well as
+    useful mathematical operations for 2D geometry and linear algebra.
+
+    :py:class:`Vector2` implements :py:class:`Sequence
+    <collections.abc.Sequence>` (and subsequently :py:class:`Iterable
+    <collections.abc.Iterable>`), allowing usage like converting, indexing, and unpacking:
```
astronouth7303: I left these off deliberately, especially `Sequence` because `Vector2` is not technically a `collections.abc.Sequence` because it does not implement several of the functions.

It's just an informal sequence because it's iterable, indexable, and has a length.

nbraud: Ah, ok.

nbraud: @astronouth7303 BTW, why aren't we inheriting from `collections.abc.Sequence`?

astronouth7303: I don't remember if there was a particular reason, or just "we only kinda pretend to be a sequence, but it's more like a struct/bag", or just didn't bother?

nbraud: Should we inherit from it?  Being “kinda sort a sequence but not really” sounds like there are some fun edge-cases waiting to bite someone in the butt.

nbraud: (Definitely not something for this PR though)

astronouth7303: The unimplemented stuff is kinda weird on a Vector? Especially the search stuff.

astronouth7303: Like, if you look at the [methods associated with `Sequence`](https://docs.python.org/3/library/collections.abc.html), many of them vary from "ehhhhh" to "just kinda not applicable?".

nbraud:
@astronouth7303 What do you think of https://github.com/ppb/ppb-vector/pull/130/commits/d51a958c72ff9d5f6bea170842ec9974ec5a508c

astronouth7303:
Ehhh... I'm not sure it's less awkward, but it's less-complicated terms? idk.

nbraud:
@astronouth7303 Part of the point is that not all methods return `Vector`, so I wanted to clarify that, but I also think it's still an awkward formulation.  :(

nbraud:
@pathunstrom Do you think you will have time to review this, or should @astronouth7303 and I go ahead and merge?

docs/conf.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'ppb-vector'
+copyright = '2019, Piper Thunstrom, Jamie Bliss'
+author = 'Piper Thunstrom, Jamie Bliss'
+
+# The short X.Y version
+version = ''
```
pathunstrom: Is this autofilled?

astronouth7303: Not as far as I know? I'm also not sure where it's used.

nbraud:
bors r=pathunstrom

nbraud:
@astronouth7303 Looks like Travis CI picked up on Bors' temporary branch, failed to clone them due to a race-condition (Bors deleted the temp branch before Travis could clone it), and marked the PR as failing as a result; do you think this is worth fixing?

pathunstrom:
Didn't we try to keep Travis from running against the tmp branches?

nbraud:
@pathunstrom Nope, I forgot about Travis in #135 ;_;

nbraud:
This took a while because the Windows tasks all timed out waiting to start (after 15 minutes) the first time around; I guess Cirrus is having issues with their Windows boxes?

pathunstrom:
Better it wastes time on an unwatched server than one of our machines.

I think I like this bors process, even if I've not initiated it yet.

nbraud:
Yes, I find it pretty convenient, since one can just tell it to merge and it will notify once CI is done on the merge commit, and push the merge (if CI is green), rather than having a human waiting on CI  :)

PR #127, opened 2019-03-23
Remove Travis CI

nbraud:
So long, and thanks for all the fish.

To be merged once we feel confident Cirrus CI is working as desired.

PR #141, opened 2019-04-02
setup.py improvements

nbraud:
Mostly the same thing as ppb/pursuedpybear#223 and ppb/pursuedpybear#224 :

- [x] Rename `*-requirements.txt` to `requirements-*.txt` to follow PPB's convention.
- [x] Extract runtime dependencies to `requirements.txt`.
- [x] Move all metadata from `setup.py` to `setup.cfg`.
- [x] Declare test dependencies in `setup.py`, integrate with `pytest`.
- [x] Use `pyproject.toml` to explicitly declare that we are using `setuptools`.


nbraud:
I ran into a Travis issue, which I fixed in the latest commit here.

I'd honestly prefer to remove that commit and do away with Travis (by merging #127):
It's been almost 2 weeks and 13 pull requests since we set up Cirrus, and while the situation was different earlier (because Cirrus was newer and we didn't need to touch Travis for it to keep working), I'm not keen on maintaining 2 parallel CIs.

nbraud:
OK, this is ready for review and merge; I removed the Travis fixes that would have conflicted with #127

nbraud:
PS: I chose not to make `setup.py test` run the linters for two reasons:

0. A new version of the linter may add new warnings, but that's no reason to fail `setup.py test && setup.py install`, especially since most Linux distros use something equivalent when packaging Python things.
1. It would make the linters a test dependency (hence, a build dependency in contexts where packages are tested at build time).

astronouth7303:
lgtm
bors r+

PR #142, opened 2019-04-04
Move to a non-copying constructor

### LINKED ISSUE #137: Implement __new__ to avoid needless copies
See discussion in #132

astronouth7303:
Also remove `.convert()`.

Closes #137

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -208,10 +208,10 @@ class Vector2:
         """
         rtype = _find_lowest_vector(type(other), type(self))
         try:
-            otherx, othery = Vector2._unpack(other)
+            other_x, other_y = Vector2._unpack(other)
         except ValueError:
             return NotImplemented
-        return rtype(self.x + otherx, self.y + othery)
+        return rtype(self.x + other_x, self.y + other_y)
 
     def __sub__(self: Vector, other: VectorLike) -> Vector:
         """Subtract one vector from another.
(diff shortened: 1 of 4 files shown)


ppb_vector/vector2.py
```diff
@@ -116,10 +129,17 @@ def __init__(self, *args, **kwargs):
         if kwargs:
             x, y = kwargs['x'], kwargs['y']
         elif len(args) == 1:
-            x, y = Vector2.convert(args[0])
+            value = args[0]
+            if isinstance(value, cls):
+                # Short circuit if already a valid instance
+                return value
+
+            x, y = Vector2._unpack(args[0])
```
nbraud:
```suggestion
            x, y = Vector2._unpack(value)
```

ppb_vector/vector2.py
```diff
@@ -198,42 +201,42 @@ def __add__(self: Vector, other: VectorLike) -> Vector:
...
-          For a description of vector-likes, see :py:func:`convert`.
+          For a description of vector-likes, see :py:func:`__new__`.
 
         >>> Vector2(1, 0) + (0, 1)
         Vector2(1.0, 1.0)
         """
         rtype = _find_lowest_vector(type(other), type(self))
         try:
-            other = Vector2.convert(other)
+            otherx, othery = Vector2._unpack(other)
```
nbraud:
```suggestion
            other_x, other_y = Vector2._unpack(other)
```

I think that would be easier to read; of course, that needs to be fixed thorough the code, but that's nothing a search-and-replace cannot handle. I don't have much opinions on snake_case vs. camelCase, but I think the former is customary in Python?)

PR #144, opened 2019-04-06
Update expectations of subclasses

nbraud:
- [x] Document status quo.
- [x] Propose support of subclasses that define new attributes, described in documentation and tests.
  Those are required to be `dataclasses`, to implement `__new__`, and not to implement `__init__`.
- [x] Implement that support using `update()`
  Typing guarantees are weakened: returned vectors are now guaranteed to have the type of `self`.
- [x] Make the references friendlier, in the inheritance docs.
- [x] Measure whether there is a performance impact.
- [ ] Fix the performance regression.
- [ ] Fix the issue with `__new__` (see #146).

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -1,5 +1,4 @@
 import dataclasses
-import functools
 import typing
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
@@ -18,35 +17,6 @@ VectorLike = typing.Union[
     typing.Sequence[typing.SupportsFloat],  # TODO: Length 2
     typing.Mapping[str, typing.SupportsFloat],  # TODO: Length 2, keys 'x', 'y'
 ]
-
-
-@functools.lru_cache()
-def _find_lowest_type(left: typing.Type, right: typing.Type) -> typing.Type:
-    """
-    Guess which is the more specific type.
-    """
-    # Basically, see what classes are unique in each type's MRO and return who
-    # has the most.
-    lmro = set(left.__mro__)
-    rmro = set(right.__mro__)
-    if len(lmro) > len(rmro):
-        return left
-    elif len(rmro) > len(lmro):
-        return right
-    else:
-        # They're equal, just arbitrarily pick one
-        return left
-
-
-def _find_lowest_vector(left: typing.Type, right: typing.Type) -> typing.Type:
-    if left is right:
-        return left
-    elif not issubclass(left, Vector2):
-        return right
-    elif not issubclass(right, Vector2):
-        return left
-    else:
-        return _find_lowest_type(left, right)
 
 
 @dataclass(eq=False, frozen=True, init=False, repr=False)
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -209,12 +209,12 @@ class Vector2:
         >>> Vector2(1, 0) + (0, 1)
         Vector2(1.0, 1.0)
         """
-        rtype = _find_lowest_vector(type(other), type(self))
         try:
             other_x, other_y = Vector2._unpack(other)
         except ValueError:
             return NotImplemented
-        return rtype(self.x + other_x, self.y + other_y)
+
+        return self.update(x=self.x + other_x, y=self.y + other_y)
 
     def __sub__(self: Vector, other: VectorLike) -> Vector:
         """Subtract one vector from another.
@@ -225,12 +225,12 @@ class Vector2:
         >>> Vector2(3, 3) - (1, 1)
         Vector2(2.0, 2.0)
         """
-        rtype = _find_lowest_vector(type(other), type(self))
         try:
             other_x, other_y = Vector2._unpack(other)
         except ValueError:
             return NotImplemented
-        return rtype(self.x - other_x, self.y - other_y)
+
+        return self.update(x=self.x - other_x, y=self.y - other_y)
 
     def dot(self: Vector, other: VectorLike) -> float:
         """Dot product of two vectors.
@@ -253,7 +253,7 @@ class Vector2:
         Vector2(3.0, 6.0)
         """
         scalar = float(scalar)
-        return type(self)(scalar * self.x, scalar * self.y)
+        return self.update(x=scalar * self.x, y=scalar * self.y)
 
     @typing.overload
     def __mul__(self: Vector, other: VectorLike) -> float: pass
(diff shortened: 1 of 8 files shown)


nbraud:
Lints are currently failing with `mypy` complaining about (the uses of) the constructor for `LabeledVector`, in `tests/subclass`.

nbraud:
Just to clarify, I made this so we could have a concrete proposal to discuss and iterate on.
I don't necessarily expect what gets eventually merged to be close to my initial proposal.

nbraud:
I “fixed” the mypy issue this morning (with `type: ignore` annotations...) but I wouldn't mind knowing how to properly do this  :)

nbraud:
So, any comments about the suggested change?

nbraud:
I finally took the time to run the benchmarks, and the results seem pretty clear: there's a large performance hit for all methods that return a vector:
```
$ python3 -m perf compare_to tests/benchmark_*.json --table                                                                             
+-----------+------------------+-------------------------------+
| Benchmark | benchmark_master | benchmark_subclass            |
+===========+==================+===============================+
| __add__   | 2.88 us          | 5.85 us: 2.03x slower (+103%) |
+-----------+------------------+-------------------------------+
| __sub__   | 2.92 us          | 5.88 us: 2.01x slower (+101%) |
+-----------+------------------+-------------------------------+
| reflect   | 9.46 us          | 16.2 us: 1.72x slower (+72%)  |
+-----------+------------------+-------------------------------+
| dot       | 618 ns           | 595 ns: 1.04x faster (-4%)    |
+-----------+------------------+-------------------------------+
| isclose   | 7.83 us          | 11.5 us: 1.47x slower (+47%)  |
+-----------+------------------+-------------------------------+
| __neg__   | 2.50 us          | 5.61 us: 2.24x slower (+124%) |
+-----------+------------------+-------------------------------+
| normalize | 6.46 us          | 13.3 us: 2.06x slower (+106%) |
+-----------+------------------+-------------------------------+
| rotate    | 4.06 us          | 7.69 us: 1.89x slower (+89%)  |
+-----------+------------------+-------------------------------+
| scale_by  | 2.23 us          | 5.39 us: 2.41x slower (+141%) |
+-----------+------------------+-------------------------------+
| scale_to  | 6.28 us          | 13.1 us: 2.09x slower (+109%) |
+-----------+------------------+-------------------------------+

Not significant (5): angle; __eq__; Vector2; length; truncate
```

Would there be a way to either:
1. have a faster, specialized version of `Vector2.update` (that can only deal with modifying `x` and `y`),
2. or make `Vector2.update` faster?

astronouth7303:
`Vector2.update()` is literally `dataclasses.replace()`. Looking at [the source](https://github.com/python/cpython/blob/ac31da8f3710f9f9b8dbb4c36b2108fb1e5b4a48/Lib/dataclasses.py#L1205-L1253), it's a single loop over the fields.

Optimizing that sounds non-trivial, and possibly involve native code.

The fact that it is significantly slower than the class-searching stuff is disconcerting, though.

nbraud:
@astronouth7303 That was my point, though: `dataclasses.update` is a completely-generic thing, that doesn't know *a priori* whichs fields are in its arguments or in the dataclass, whereas we should be able to do much better (for instance with a shallow copy of the object, then setting `x` and `y`).

nbraud:
Fixing the performance regression is blocked by #146.

astronouth7303:
Except that we don't not what fields there will be? That's the whole point of this whole thing?

nbraud:
@astronouth7303 Yes, but we know that we only ever want to modify `x` and `y`, and that they exist (as in, are attributes of the object), so if we can do a quick shallow-copy and then update `x` and `y`, that's fine.

ppb_vector/vector2.py
```diff
@@ -98,18 +107,19 @@ def __new__(cls, *args, **kwargs):
...
             x, y = kwargs['x'], kwargs['y']
+
         elif len(args) == 1:
             value = args[0]
-            if isinstance(value, cls):
-                # Short circuit if already a valid instance
-                return value
+            # Early return if the argument is a vector:
+            #  __new__ returned the same vector, no need to (re)set x and y
+            if isinstance(args[0], type(self)):
```
astronouth7303: You called me out once for using `args[0]` instead of `value`

tests/test_subclass.py
```diff
@@ -12,10 +12,9 @@ class LabeledVector(Vector2):
     """Subclass of Vector2 that defines an additional attribute."""
     label: str
 
-    def __new__(cls, x, y, label):
-        self = super().__new__(cls, x, y)
+    def __init__(self, x, y, label):
+        Vector2.__init__(self, x, y)
```
astronouth7303: `super()`

tests/test_subclass.py
```diff
@@ -75,7 +74,7 @@ def test_subclass_scalar(op, v: Vector2, label: str, scalar: float):
     assert u.label == label
 
 
-@pytest.mark.parametrize("op", UNARY_OPS)
+@pytest.mark.parametrize("op", UNARY_OPS - set([Vector2]))
```
astronouth7303: `set([Vector2])` == `{Vector2}`

ppb_vector/vector2.py
nbraud: And you are entirely right to do that here.  I think it's because it got copy-pasted from `__new__`, where there's no variable `value`.

tests/test_subclass.py
nbraud: Those 2 don't seem equivalent; you can check, all tests in that file fail if you make it `super()`.

ppb_vector/vector2.py
nbraud: Oh, this came from #146, so I fixed it there too.

tests/test_subclass.py
nbraud: Nevermind, I'm just bad at code

nbraud:
> I'd still prefer a .__copy__() that just returns self.

@astronouth7303 How is this related to (the current state of) this PR?
We can implement `__copy__` later, assuming `update` (or some specialized version of it) doesn't do `copy.copy` and in-place mutation (but that would be unlikely, since `copy.copy` uses the pickling protocol, which is quite slow)

tests/test_subclass.py
pathunstrom: So, calling the base class explicitly like this is actually an older (and less accepted now) pattern in Python. What was the problem you were trying to solve?

nbraud: Yeah, this come from #146 where I had initially used `super()`, but messed it up and ended up calling directly so it would work.  @astronouth7303 already commented on that and it already got fixed.

nbraud: See https://github.com/ppb/ppb-vector/pull/144#discussion_r275822548

PR #146, opened 2019-04-11
Vector2.__new__ violates its interface contract

nbraud:
While working on #144, I discovered that instances of `Vector2` cannot be copied (using `copy.{copy,deepcopy}`), and I believe this is because we are violating the interface contract for `__new__`

- [x] Provide a test reproducing the issue.
- [x] Fix the issue.
- [x] Update #144 so we do not have broken examples and expectations in the documentation.

### DIFF
--- a/ppb_vector/__init__.py
+++ b/ppb_vector/__init__.py
@@ -83,6 +83,15 @@ class Vector:
 
         - any instance of :py:class:`Vector` or any subclass.
         """
+        if args and len(args) == 1:
+            value = args[0]
+            if isinstance(value, cls):
+                # Short circuit when a valid instance is provided
+                return value
+
+        return super().__new__(cls)
+
+    def __init__(self, *args, **kwargs):
         if args and kwargs:
             raise TypeError("Got a mix of positional and keyword arguments")
 
@@ -95,17 +104,18 @@ class Vector:
 
         if kwargs:
             x, y = kwargs['x'], kwargs['y']
+
         elif len(args) == 1:
             value = args[0]
-            if isinstance(value, cls):
-                # Short circuit if already a valid instance
-                return value
+            # Early return if the argument is a vector:
+            #  __new__ returned the same vector, no need to (re)set x and y
+            if isinstance(value, type(self)):
+                return
 
             x, y = Vector._unpack(value)
+
         elif len(args) == 2:
             x, y = args
-
-        self = super().__new__(cls)
 
         try:
             # The @dataclass decorator made the class frozen, so we need to
@@ -120,8 +130,6 @@ class Vector:
             object.__setattr__(self, 'y', float(y))
         except ValueError:
             raise TypeError(f"{type(y).__name__} object not convertable to float")
-
-        return self
 
     def __reduce__(self):
         return Vector.__new__, (Vector, self.x, self.y)


nbraud:
@astronouth7303 Assigning this to you because I do not know well-enough the internals of the Python object model  :3

nbraud:
According to [`__new__`'s spec]. `__init__` is called automatically afterwards, so I moved more of the logic back in it (with the added benefit that subclasses will get the non-copying behaviour “for free”).
It looks like this resulted in some progress:
```py3
>>> import copy
>>> copy.copy(Vector2(0, 0))                                                                 
---------------------------------------------------------------------------
FrozenInstanceError                       Traceback (most recent call last)
<ipython-input-6-7fbabe237cc4> in <module>
----> 1 copy.copy(Vector2(0, 0))

/usr/lib/python3.7/copy.py in copy(x)
    104     if isinstance(rv, str):
    105         return x
--> 106     return _reconstruct(x, None, *rv)
    107 
    108 

/usr/lib/python3.7/copy.py in _reconstruct(x, memo, func, args, state, listiter, dictiter, deepcopy)
    290             if slotstate is not None:
    291                 for key, value in slotstate.items():
--> 292                     setattr(y, key, value)
    293 
    294     if listiter is not None:

<string> in __setattr__(self, name, value)

FrozenInstanceError: cannot assign to field 'x'
```

Clearly something is still broken, though. I checked, and regular dataclasses are copyable without issue:
```py3
>>> from dataclasses import dataclass                                                        

>>> @dataclass(frozen=True) 
>>> class Foo: 
>>>     bar: str 
>>>     weight: float 

>>> copy.copy(Foo("blah", 0.5))                                                              
Foo(bar='blah', weight=0.5)
```

[`__new__`'s spec]: https://docs.python.org/3/reference/datamodel.html#object.__new__

ppb_vector/vector2.py
```diff
@@ -154,10 +164,8 @@ def __new__(cls, *args, **kwargs):
         except ValueError:
             raise TypeError(f"{type(y).__name__} object not convertable to float")
 
-        return self
-
     def __reduce__(self):
-        return type(self).__new__, (type(self), self.x, self.y)
+        return type(self), (self.x, self.y)
```
nbraud: Using `type(self).__new__` there violates the protocol for [object construction]:

> If `__new__` returns an instance of `cls`, then the new instance’s `__init__()` method will be invoked like `__init__(self[, ...])`, where `self` is the new instance and the remaining arguments are the same as were passed to `__new__`.

[object construction]: https://docs.python.org/3/reference/datamodel.html#object.__new__

nbraud:
*pings @astronouth7303*

I ended up merging this branch into #144, because I ended up running into some of the issues fixed here, but I'd prefer to get this reviewed and merged first.

The specific issue with `pickle` and `copy` was fixed when you implemented `__reduce__`, but I think protocol violations are worth fixing (especially since this one manifested itself as hard-to-debug wonkyness).

nbraud:
Rebased to deal with a merge conflict.

astronouth7303:
I'm just not buying that an argumentless `__new__` is part of the contract.

nbraud:
How come that `pickle` (and possibly other modules) make that assumption, then?

I guess I should split this into several PRs, and only keep the controversial `__new__`/`__init__` change here.

ppb_vector/vector2.py
nbraud: Extracted to #157.

nbraud:
PS: Since I extracted the fix for `__reduce__` to #157, this fails in hilarious ways.
I will pull #157 into this branch, but please merge it first.

astronouth7303:
It appears that when a custom `__reduce__`/`__reduce_ex__`/etc (there's like 5 possible dunder methods?) is not defined, it defaults to saving the `__dict__`, calling argless `__new__`, and restoring the `__dict__`.

This is fine for many standard Python classes (eg, the engine, systems, scenes, and sprites), this assumption falls apart for a slots-based immutable class.

Keep in mind that pickle is standard library, not core interpreter. It's doing a lot of magic, but is making some assumptions while doing so.

astronouth7303:
Going to close this.

PR #147, opened 2019-04-12
Pickle support

astronouth7303:
Add support for pickling.

### DIFF
--- a/ppb_vector/vector2.py
+++ b/ppb_vector/vector2.py
@@ -156,6 +156,9 @@ class Vector2:
 
         return self
 
+    def __reduce__(self):
+        return type(self).__new__, (type(self), self.x, self.y)
+
     #: Return a new :py:class:`Vector2` replacing specified fields with new values.
     update = dataclasses.replace
 
--- a/tests/test_vector2_ctor.py
+++ b/tests/test_vector2_ctor.py
@@ -44,4 +44,4 @@ def test_ctor_noncopy_superclass():
 @given(v=vectors())
 def test_ctor_copy(v: Vector2):
     """Test that Vector2 instances can be copied."""
-    assert v == pickle.dumps(pickle.loads(v))
+    assert v == pickle.loads(pickle.dumps(v))
--- a/tests/test_vector2_ctor.py
+++ b/tests/test_vector2_ctor.py
@@ -44,5 +44,5 @@ def test_ctor_noncopy_superclass():
 
 @given(v=vectors())
 def test_ctor_pickle(v: Vector2):
-    """Test that Vector2 instances can be pickled."""
+    """Round-trip vectors through `pickle.{dumps,loads}`."""
     assert v == pickle.loads(pickle.dumps(v))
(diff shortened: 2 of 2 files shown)


ppb_vector/vector2.py
```diff
@@ -156,6 +156,9 @@ def __new__(cls, *args, **kwargs):
 
         return self
 
+    def __reduce__(self):
+        return type(self).__new__, (type(self), self.x, self.y)
```
nbraud: Will that prevent subclasses (esp. ones that add attributes) to be pickled correctly?

tests/test_vector2_ctor.py
```diff
@@ -38,3 +40,9 @@ def test_ctor_noncopy_subclass():
 def test_ctor_noncopy_superclass():
     v = Vector2(1, 2)
     assert V(v) is not v
+
+
+@given(v=vectors())
+def test_ctor_copy(v: Vector2):
+    """Test that Vector2 instances can be copied."""
```
nbraud: Docstring and function name need to reflect what the test does.

Also, is `test_vector2_ctor.py` the right location for tests of a protocol that isn't the constructor?

tests/test_vector2_ctor.py
```diff
@@ -38,3 +40,9 @@ def test_ctor_noncopy_subclass():
 def test_ctor_noncopy_superclass():
     v = Vector2(1, 2)
     assert V(v) is not v
+
+
+@given(v=vectors())
+def test_ctor_pickle(v: Vector2):
+    """Test that Vector2 instances can be pickled."""
```
nbraud:
```suggestion
    """Round-trip vectors through `pickle.{dumps,loads}`."""
```

We aren't just testing that they can be pickled, but that unpickling gives the correct result.

tests/test_vector2_ctor.py
```diff
@@ -38,3 +40,9 @@ def test_ctor_noncopy_subclass():
 def test_ctor_noncopy_superclass():
     v = Vector2(1, 2)
     assert V(v) is not v
+
+
+@given(v=vectors())
+def test_ctor_pickle(v: Vector2):
```
nbraud: This should likely take a `cls` argument and be parametrised over `Vector2` and a dummy subclass of it, to check we correctly (un)pickle derived classes.

tests/test_vector2_ctor.py
nbraud: @astronouth7303 You didn't reply, re: the choice of file for testing the pickle protocol.

astronouth7303: It has more to do with object instantiation than any operation.

nbraud: M'kay.

ppb_vector/vector2.py
nbraud: The improved test answers this positively, in the case of subclasses without extra attributes.

I guess it would make sense to punt the with-attributes case to #144.

astronouth7303: The pickle docs mention higher-level protocols, but they're not actually implemented by the core stuff? not super clear.

PR #148, opened 2019-04-12
Cut a 1.0 Alpha 1 release

astronouth7303:
Getting a pre-release onto PyPI allows us to unblock PPB and more easily get feedback from any testers.

Note that this is not expected to be final, and there are clearly issues we want to resolve before 1.0 final.

However, the basic API seems to have stabilized, so other projects (PPB) can move forward against this API.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = ppb-vector
-version = 0.4.0rc1
+version = 1.0a1
 author = Piper Thunstrom
 author_email = pathunstrom@gmail.com
 description = A basic game development Vector2 class.


astronouth7303:
I can do it.

bors r+

astronouth7303:
bors retry

nbraud:
@astronouth7303 FWIW, I think it's weird to claim API stabilisation when we have a bunch of API-impacting bugs that aren't resolved yet:
- #146 and #147: work with `copy` and `pickle`, resolve the `__new__` weirdness;
- #144: define what's expected of subclasses.

That said, I don't think this should hold off an alpha release, esp. since this is blocking for ppb itself.

astronouth7303:
I meant the core usage stuff.

pathunstrom:
Yeah, the only big change to the external API is the immutability, and I don't see that external API changing much going forward.

PR #150, opened 2019-05-09
[ci] fixed allow failures config

### LINKED ISSUE #134: Cirrus CI: The allow_failure parameter is incorrect
Looks like `$CIRRUS_TASK_NAME` doesn't include the matrix build modifier, so it's just `Linux` or `Windows`.  As such, setting `allow_failures: $CIRRUS_TASK_NAME =~ '.*-rc-.*'` doesn't work (failures are never allowed).

Unfortunately, Cirrus does not document an environment variable that does depend on container image's name (or anything else we could use to distinguish the CPython 3.8-rc builds).

fkorotkov:
Fixes #134


nbraud:
Looks like @bors[bot] doesn't accept wildcards in `status`, so I'd have to put in the full check's name (`${OS} IMAGE:${image} container:${image}`) which isn't the best  :(

@ppb/maintainers Any opinions on whether I should Just Do It, since we don't often mess with `bors.toml`?

pathunstrom:
I'm okay with a little manual process, and as a rule am okay with explicit configuration. We can find a way to automate updating it via scripts later? (In other words, go ahead and _just do it_ and we'll deal with it better later.)

nbraud:
@fkorotkov Thanks, I had somehow forgotten to account for your latest commit there.
This is most excellent, thanks again for the help  :purple_heart: 

bors r+

.cirrus.yml
```diff
@@ -51,15 +51,18 @@ FreeBSD_task:
...
+  env:
     matrix:
-      - image: python:3.6-slim
-      - image: python:3.7-slim
-      - image: python:3.8-rc-slim
-      - image: pypy:3.6-slim
+      - IMAGE_TAG: 3.6-slim
+      - IMAGE_TAG: 3.7-slim
+      - IMAGE_TAG: 3.8-rc-slim
+      - IMAGE_TAG: 3.6-slim
```
nbraud: Dropping the image name created some confusion between `python:3.6-slim` and `pypy:3.6-slim`.
(The former uses the CPython interpreter, the latter uses the PyPy JIT-compiled VM)

Fixing that then merging.

nbraud:
@fkorotkov Sorry for the churn on that PR, and thanks, again, for the help  :)

fkorotkov:
@nbraud NP, happy to assist! 🙌

PR #153, opened 2019-05-23
tests/ctor: Make test_ctor_noncopy_same an Hypothesis test

nbraud:
This likely doesn't matter all that much, but avoids hardcoding an arbitrary value.


astronouth7303:
I feel like this is testing an optimization, not a reliable behavior? But it's fine.

bors r+

nbraud:
@astronouth7303 You are the one who added the test (in fbf3fee245bc20a2a404367e2f395a0c54d9903b)  :P

astronouth7303:
Oh.

PR #156, opened 2019-05-24
tests/utils: Freeze all the operators sets

nbraud:
Split off from #146.

### DIFF
--- a/tests/benchmark.py
+++ b/tests/benchmark.py
@@ -9,10 +9,10 @@ x = Vector(1, 1)
 y = Vector(0, 1)
 scalar = 123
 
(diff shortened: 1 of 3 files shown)


nbraud:
Fails with a `mypy` error:
```
tests/benchmark.py:19: error: <nothing> has no attribute "__name__"
```

This sounds possibly like a bug?  :/

astronouth7303:
Is there a reason for the change? Or just most correct semantically?

nbraud:
I accidentally mutated one of those from a test, which of course had all sorts of hilarious results.
So, yes, semantic correctness and removing feetguns.

PR #158, opened 2019-05-25
Test scalar multiplication/division with int and zero

nbraud:


tests/test_scalar_multiplication.py
```diff
@@ -50,3 +51,23 @@ def test_scalar_inverse(x: Vector, scalar: float):
...
+    assume(scalar != 0)
+    assert x / scalar == x / float(scalar)
+
+
+@given(x=vectors())
+def test_division_by_zero(x: Vector):
+    with raises(ZeroDivisionError):
+        x / 0
+
+    with raises(ZeroDivisionError):
```
astronouth7303: Why twice?

tests/test_scalar_multiplication.py
```diff
@@ -1,6 +1,7 @@
 from math import isclose
 
-from hypothesis import assume, given
+from hypothesis import assume, given, strategies as st
+from pytest import raises  # type: ignore
```
astronouth7303: :sigh: what complains about this?

tests/test_scalar_multiplication.py
nbraud: With integer `0` and floating-point `0.`:
```
>>> type(0), type(0.)
(int, float)
```

tests/test_scalar_multiplication.py
nbraud: `mypy` simply doesn't have an interface definition for `pytest`, which is why all `pytest` imports need to have a `type: ignore` annotation.

tests/test_scalar_multiplication.py
astronouth7303: Oh! Missed the dot.

PR #164, opened 2019-05-31
Test the trigonometry routine on remarkable angles

nbraud:
- [x] Compare the remarkable angles to `Vector._trig`, and not `cos, sin`, in `test_remarkable_angles`.
- [x] Tighten-up the tolerances on that test.
- [x] Modify `utils.isclose` to be usable outside Hypothesis.
- [x] Modify `rotate/remarkable_angles` to use the same convention as `Vector._trig`.

### DIFF
--- a/tests/test_rotate.py
+++ b/tests/test_rotate.py
@@ -28,34 +28,34 @@ def test_exact_rotations(input, angle, expected):
 #  values from 0 to 45°
 #  lifted from https://en.wikibooks.org/wiki/Trigonometry/Selected_Angles_Reference
 remarkable_angles = {
-    15: ((sqrt(6) - sqrt(2)) / 4, (sqrt(6) + sqrt(2)) / 4),
-    22.5: (sqrt(2 - sqrt(2)) / 2, sqrt(2 + sqrt(2)) / 2),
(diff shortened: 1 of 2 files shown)


nbraud:
I mentioned on Discord that `Vector._trig` was producing values that were off (when compared to `remarkable_angles`).  The initial test was actually buggy: `remarkable_angles` contained `(sin, cos)` pairs, whereas `Vector._trig` returns `(cos, sin)` pairs.

astronouth7303:
I should note that if we have imprecision/inconsistency in these angles, I think we should take the answer that gives us smooth curves than what's 100% correct.

bors r+

nbraud:
@astronouth7303 What do you mean by “smooth curves“ here ?

In any case, rotations should be “smooth” for whatever reasonable meaning of it; I can probably write a test to check they are [1-Lipschitz].

[1-lipschitz]: https://en.wikipedia.org/wiki/Lipschitz_continuity

astronouth7303:
I just mean, I don't want any perturbances or "bumps" or jitter caused by certain angles being hardcoded or whatever.

nbraud:
@astronouth7303 Oh, yes; only the test uses known values, not the implementation.

PR #149, opened 2019-05-09
Drop subclassing support

nbraud:
- [x] Document that subclassing isn't supported.
- [x] Drop the subclassing testsuite.
- [x] Drop the subclassing support.
- [x] Rename `Vector2` to `Vector`.
  Avoid breaking backwards-compatibility by adding a `Vector2` alias.
- [x] Rename `vector2.py` to `vector.py`.

Closes #144 (this is an alternative proposal)

PR #167, opened 2019-06-01
Improve Vector.update

nbraud:
Following #149, `Vector.update` can be made simpler and faster.


nbraud:
This came up while I was adding `Vector.zero` as a class-level variable, which somehow made `dataclasses.update` unhappy.

PR #168, opened 2019-06-01
Add Vector.zero as a class-level attribute

nbraud:


nbraud:
Looks like the version of `dataclasses` in PyPI, which we use with Py3.6, doesn't handle properly `typing.ClassVar`.

astronouth7303:
I guess comment out the annotation, with a link to that bug report.

nbraud:
@astronouth7303 Without the annotation, `dataclasses` assumes it is an instance-level attribute and explodes.

I ported the fix from CPython to PyPI's `dataclasses`: https://github.com/ericvsmith/dataclasses/pull/145

nbraud:
Update: There already was a PR that backported that fix and another: https://github.com/ericvsmith/dataclasses/pull/140
Unfortunately, it's been sitting there since December, so we can't expect upstream to merge that and cut a release any time soon.  :(

astronouth7303:
Ok, so what parts of dataclasses exactly chokes on this?

If we can't find a workaround, lets turn this into a factory method.

nbraud:
Yeah, a factory method would work, but I'm willing to let this sit until there's a fixed version of `dataclasses` (or until we decide only to support Py3.7 and 3.8, whichever comes first) since it's really not high priority.

nbraud:
bors try

nbraud:
One macOS task seems to not have run.  Rude  :(

In any case, that confirms the testsuite passes now that #167 has been merged, though I would hold this up until either `dataclasses` is fixed or we drop Python 3.6 support.

nicoonoclaste:
@AstraLuma Rebased to fix the merge conflict, should be *finally* good to go.

AstraLuma:
This breaks on 3.6, right?

nicoonoclaste:
> This breaks on 3.6, right?

Not sure what you mean there, but:
- I merged #190 so 3.6 support is gone; presumably, if it was OK to go 2 years ago, it was still OK now.
- Even if we were to revert that, this code is fine on Py3.6; however, the unmaintained `dataclasses` compatibility shim for Py3.6 has (or had?) bugs that made it explode on `ClassVar` annotations, so for instance `dataclasses.replace` would choke on a `Vector` (IIRC throw an exception)

PR #170, opened 2019-06-06
Fix CI releases

astronouth7303:
Get CI release builds working.

### DIFF
--- a/.ci/upload-build.sh
+++ b/.ci/upload-build.sh
@@ -23,11 +23,15 @@ for fpath in "${files_to_upload[@]}"; do
 	echo "Uploading '$fpath' to Github..."
 	name=$(basename "$fpath")
 	url_to_upload="https://uploads.github.com/repos/$CIRRUS_REPO_FULL_NAME/releases/$CIRRUS_RELEASE/assets?name=$name"
-	curl -X POST \
+	curl -i -X POST \
 		--data-binary @$fpath \
 		--header "Authorization: token $GITHUB_TOKEN" \
 		--header "Content-Type: $file_content_type" \
 		$url_to_upload
 done
 
-twine upload "${files_to_upload[@]}"
+twine upload \
+	--repository-url "$TWINE_REPOSITORY_URL" \
+	--username "$TWINE_USER" \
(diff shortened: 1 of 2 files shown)


astronouth7303:
I'm not a fan of the upload task running on every build, but it doesn't actually upload unless it's a real release, so :shrug:.

This seems to work in my [test repo](https://github.com/astronouth7303/cirrus-release-test)

nbraud:
Rebased to get rid of the env task definition; I'd rather not have that at all in the history, given that executing it (from the ppb-vector repo) would cause Cirrus to dump its secrets.

nbraud:
Here is an example of the `upload` task running in the test repository: https://cirrus-ci.com/task/5429082557251584

astronouth7303:
Upload task attached to a release: https://cirrus-ci.com/task/5176411040514048

The env task should be fine, given secrets are attached to specific tasks.

.ci/upload-build.sh
```diff
@@ -20,14 +20,22 @@ file_content_type="application/octet-stream"
...
 	echo "Uploading '$fpath' to Github..."
 	name=$(basename "$fpath")
 	url_to_upload="https://uploads.github.com/repos/$CIRRUS_REPO_FULL_NAME/releases/$CIRRUS_RELEASE/assets?name=$name"
-	curl -X POST \
+	curl -i -X POST \
 		--data-binary @$fpath \
 		--header "Authorization: token $GITHUB_TOKEN" \
 		--header "Content-Type: $file_content_type" \
 		$url_to_upload
+	echo
```
nbraud: Please remove that `echo`, we are already inserting a linebreak at the top of the loop or right below.

astronouth7303: This echo is for the curl. The one above is for a stylistic line break so that the various parts are visually broken up.

It's called ux.

astronouth7303: "This line deliberately left blank"

nbraud: > It's called ux

No need for snark...
It was entirely unobvious that `curl` fails to insert a linebreak before returning.

PR #173, opened 2019-06-09
Minor docs improvements

nbraud:
- Document `__bool__`.
- Document in `__add__` and `__mul__` that the reverse operator (`vector-like op vector`) also exists.
- Use the imperative form in short-form descriptions (i.e. the docstring's first line).
- Use asserts in examples where relevant, rather than showing `True` or implicitly expecting users to notice an equality.
- Other minor changes.


astronouth7303:
bors retry

PR #178, opened 2019-07-04
Cirrus CI: Check that the PEP517 spec works and builds sensibly

nbraud:
Actually catches in CI the issue fixed in #177.


nbraud:
The `pep517` failure is expected until #177 is merged  :)

nbraud:
bors try

PR #179, opened 2019-08-31
tests/rotate: Add metamorphic stability test for Vector._trig

nbraud:
Assert that `Vector._trig` is closer to the unit circle than `math.{cos,sin}`.


nbraud:
Thanks to @astronouth7303 for suggesting a while back to look at metamorphic tests.

PR #180, opened 2019-08-31
tests/normalize: Test that direction is preserved

nbraud:


nbraud:
Sorry, missed the missing import when extracting this commit from another branch.

PR #181, opened 2019-08-31
Deprecate Vector.scale

nbraud:
`scale_to` is the preferred method name, as it isn't ambiguous.
@astronouth7303 added it in October 2018, so it has been available for a while now.

### DIFF
--- a/ppb_vector/__init__.py
+++ b/ppb_vector/__init__.py
@@ -487,16 +487,16 @@ class Vector:
         It produces a vector with the same direction, but possibly a different
         length.
 
-        Note that :py:meth:`vector.scale(max_length) <scale>` is equivalent to
-        :py:meth:`vector.truncate(max_length) <truncate>` when
+        Note that :py:meth:`vector.scale_to(max_length) <scale_to>` is
+        equivalent to :py:meth:`vector.truncate(max_length) <truncate>` when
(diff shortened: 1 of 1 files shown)


astronouth7303:
Why did we ship this in 1.0?

Per semver, I guess this will be removed in 2.0.

nbraud:
@astronouth7303 Eh?  Adding a deprecation warning doesn't break backwards compat.

Oh do you mean the actual removal of `Vector.scale`?

astronouth7303:
Yeah,

1. Why did we ship `.scale()` in 1.0, given we were already breaking everything?
2. When will this be removed and not just deprecated? (per semver, that's a breaking change, so 2.0)

nbraud:
> 1. Why did we ship .scale() in 1.0, given we were already breaking everything?

That would have indeed be the best time to do it, and maybe we would have noticed if there had been an explicit deprecation notice.  As it stands, I have no idea why it flew under everyone's radar, and I'm not particularly interested in assigning blame.

> 2. When will this be removed and not just deprecated? (per semver, that's a breaking change, so 2.0)

Yes, this is obviously API-breaking, so it would need a 2.0 release.

nbraud:
@astronouth7303 Ping?

astronouth7303:
Bors r+

PR #186, opened 2019-11-08
tests/rotate: Pass integer parameters to hypothesis.st.integers

nbraud:
[Hypothesis #2117], released in v4.39.3, introduced tighter type annotations, causing `mypy` to reject int-representable floats as parameters to `st.integers`.

[Hypothesis #2117]: https://github.com/HypothesisWorks/hypothesis/pull/2117

### DIFF
--- a/tests/test_rotate.py
+++ b/tests/test_rotate.py
@@ -111,7 +111,7 @@ def test_trig_stability(angle):
     assert fabs(1 - r_len) <= fabs(1 - t_len)
 
 
-@given(angle=angles(), n=st.integers(min_value=0, max_value=1e5))
+@given(angle=angles(), n=st.integers(min_value=0, max_value=100_000))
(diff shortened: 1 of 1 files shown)


tests/test_rotate.py
```diff
@@ -123,7 +123,7 @@ def test_trig_invariance(angle: float, n: int):
     assert isclose(r_sin, n_sin, rel_to=[n / 1e9])
 
 
-@given(v=vectors(), angle=angles(), n=st.integers(min_value=0, max_value=1e5))
+@given(v=vectors(), angle=angles(), n=st.integers(min_value=0, max_value=10_000))
```
astronouth7303: 10,000 is 1e4. Was this change intentional?

nbraud: Oops, no; let me PR this back, thanks for the catch.

PR #185, opened 2019-11-08
tests/benchmarks: Update for pyperf rename (from perf)

nbraud:

PR #187, opened 2019-11-08
Fix CI

nbraud:
This amalgamates the fixes from:
- #185 : Update after `perf` was renamed to `pyperf`.
- #186 : Update `tests/rotate` after Hypothesis published stricter type hints.

Moreover, this includes the following fix:
- Cirrus CI: Do not check certificate when installing the macOS SDK

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -110,7 +110,7 @@ task:
 macOS_task:
   skip: $CIRRUS_BRANCH =~ '.*\.tmp'
   osx_instance:
-    image: mojave-base
+    image: mojave-xcode-11
   env:
     PATH: ${HOME}/.pyenv/shims:${PATH}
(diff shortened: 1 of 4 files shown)


nbraud:
Had to disable the certificate check on the macOS SDK install, as that suddenly started failing.
The certificate chain embedded in the `.pkg` file probably expired.

nbraud:
CI still failing on macOS, feel free to fix as I'm likely not going to do that today.

nbraud:
@astronouth7303 It would be amazing if you could review/merge, as I'm kinda keen on working CI  :)

nbraud:
PS: As a bonus, using an image with Xcode preinstalled saves us ~1 minute (out of 4) on each macOS CI task.

PR #188, opened 2019-11-11
fixup! tests/rotate: Pass integer parameters to hypothesis.st.integers

nbraud:
1e5 is 100_000, not 10_000...

Thanks to @astronouth7303 for spotting it in review.


nbraud:
bors try

PR #182, opened 2019-11-08
CI: Update for CPython 3.8

nbraud:

PR #189, opened 2019-11-11
CI: Add CPython 3.8 environment on FreeBSD

nbraud:
Split off from #182 as `python38` was failing to install/be found on the FreeBSD CI.


astronouth7303:
This is blocked on #191

astronouth7303:
bors retry

Issue #184, opened 2019-11-08
Drop Py3.6 support?


nbraud:
Now that Python 3.8 is out, and Python 3.7 is available starting in Debian 10 “Buster” (current stable) and Ubuntu 19.04 “Disco Dingo”, should we consider dropping Python 3.6 support?

Python 3.6 support currently relies on the `dataclasses` backport, which seems essentially-unmaintained since November 2018, and lacks [important bugfixes].  Those bugs are blocking for API improvements like #168.

[important bugfixes]: https://github.com/ericvsmith/dataclasses/pull/140

astronouth7303:
This should track with PPB's own support cycle.

I think that's unofficially the last 2 major versions (so currently, 3.8 and 3.7), but I'm not sure that's been decided formally.

I'm going to just take maintainer's fiat and say that ppb-vector devel may drop 3.6. 1.0 exists and supports it, and that's good enough.

nbraud:
@astronouth7303 Yeah, that was my recollection too (last 2 major releases of CPython), but I'd rather check with you and Piper first.

nbraud:
Also, dropping Py3.6 from CI will conflict with all the CI updates that are in-flight, so that can wait a couple of days.

astronouth7303:
Revisting some stuff. Do we want to drop 3.6 given pypy?

nicoonoclaste:
> Revisting some stuff. Do we want to drop 3.6 given pypy?

PyPy has support for Py3.7 now, so I would be strongly in favour of dropping Py3.6 compatibility.
Though the `dataclasses` compat shim has been updated, so we could in principle keep support if we cared to.

pathunstrom:
I'll note, 3.5 Python just hit EOL, 3.6 is still getting security patches, we should all have a conversation about support schedule sometime.

nicoonoclaste:
> Though the `dataclasses` compat shim has been updated, so we could in principle keep support if we cared to.

Turns out, the maintainer of `dataclasses` (the PyPI package, not the stdlib module) merged the PR that contained the update, but failed to publish a new version...

nicoonoclaste:
> I'll note, 3.5 Python just hit EOL, 3.6 is still getting security patches, we should all have a conversation about support schedule sometime.

I simply don't think that we can reasonably support people on Py3.6:
[...]

PR #190, opened 2019-11-14
Drop compatiblity with Python 3.6

nbraud:
We only aim to support the last 2 major versions of Python (currently 3.7 and 3.8), and the `dataclasses` backport for Python 3.6 is unmaintained and broken (see discussion in #184).

This should allow us to unblock #168


nbraud:
I noticed a disadvantage while updating the CI config: PyPy3 recently had a stable release, but it still only supports Python 3.6.9, so we would be dropping PyPy support (until they move to Py3.7).

I don't think this is a major issue, as ppb itself doesn't support PyPy.

astronouth7303:
Not at this time, but I'd be ok pursuing pypy compatibility for performance reasons? There's other complications, though.

If the dataclasses bugs cause problems, let's drop it from ppb-vector; we're only using it for utility reasons.

nbraud:
> If the dataclasses bugs cause problems

To be clear, that's not a bug in the stdlib's `dataclasses` module (at least, not as found in any version of Python ever released).

The issue is that the (unofficial) PyPI package is essentially a snapshot from an early development version of `dataclasses` from before it was first included in a stable Python release.

nicoonoclaste:
bors r=astraluma

nicoonoclaste:
bors r=astraluma

nicoonoclaste:
OK, I guess I'm making a CI-fixing PR first  >_>'

nicoonoclaste:
bors r=astraluma

PR #191, opened 2019-12-19
Fix CI

astronouth7303:
We accidentally neglected CI to the point of multiple blocking issues that can't be resolved individually.


astronouth7303:
There's some weird CI stuff going on with FreeBSD. Sent an email to Cirrus support.

PR #193, opened 2020-05-29
Add Python 3.9 to CI

astronouth7303:
3.9 is coming.


astronouth7303:
Ok, I think this is set.

PR #195, opened 2020-10-08
CI: Upgrade to current versions of CPython and PyPy

nicoonoclaste:
- [x] Add CI environments for CPython 3.9 and 3.10-rc
- [x] Configure Bors to require CI to pass on 3.9
- [x] Add a CI environment for PyPy (w/ Python 3.7)

PR #196, opened 2020-10-08
CI: Update CPython on macOS

nicoonoclaste:


AstraLuma (approved):
lgtm

the pyenv that gets shipped on cirrus mac doesn't always track with PSF latest--I've had trouble in the past with figuring out what versions exactly are available.

AstraLuma (changes requested):
Errr... I should have looked at the test results first.

They demonstrate exactly what I'm talking about.

nicoonoclaste:
@AstraLuma I was aware that those CPython versions aren't yet available, this got split off #195 so as not to hold up the update on Linux and Windows.

nicoonoclaste:
bors try

AstraLuma:
bors d+

nicoonoclaste:
bors r=astraluma

PR #199, opened 2021-04-14
Update Python versions

AstraLuma:
Semi-annual touching of python versions.


AstraLuma:
(Thanks to @duckinator for helping on the FreeBSD issue)

AstraLuma:
@pathunstrom how's this look?

ppb_vector/__init__.py
```diff
@@ -118,12 +118,12 @@ def __new__(cls, *args, **kwargs):
             #  https://docs.python.org/3/library/dataclasses.html#frozen-instances
             object.__setattr__(self, 'x', float(x))
         except ValueError:
-            raise TypeError(f"{type(x).__name__} object not convertable to float")
+            raise TypeError(f"{type(x).__name__} object not convertable to float") from None
```
nicoonoclaste: Are we sure the lint is invalid and `from None` is the right thing here? (and same right below)

.cirrus.yml
```diff
@@ -132,13 +136,27 @@ macOS_task:
...
-    - pip install -e .
+    - pip install .
 
   script:
     - python3 --version
     - pip list
     - ./test.sh
 
+macOS_version_task:
+  skip: $CIRRUS_BRANCH =~ '.*\.tmp'
```
nicoonoclaste, line 143: We probably don't want to run that on all branches etc.

```suggestion
  skip: $CIRRUS_BRANCH =~ 'canon'
```

.cirrus.yml
```diff
@@ -94,7 +95,8 @@ task:
       - IMAGE: python:3.7-slim
       - IMAGE: python:3.8-slim
       - IMAGE: python:3.9-slim
-      - IMAGE: python:3.10-rc-slim
+      - IMAGE: python:3.10-slim
```
nicoonoclaste, line 96: Since we are using `rc-slim` below to always get the latest dev. version, why not use `3-slim` here and only use major version numbers for the older entries?

I think this would be more coherent, and has the advantage of always tracking the latest stable Python version even if we forgot to update the CI config.

```suggestion
      - IMAGE: python:3-slim
```

.cirrus.yml
```diff
@@ -147,7 +165,8 @@ task:
       - IMAGE: python:3.7-windowsservercore
       - IMAGE: python:3.8-windowsservercore
       - IMAGE: python:3.9-windowsservercore
-      - IMAGE: python:3.10-rc-windowsservercore
+      - IMAGE: python:3.10-windowsservercore
```
nicoonoclaste, line 163: Same, `3-windowsservercore`

nicoonoclaste (changes requested):
LGTM overall, but there are a few things I'd like answered or changed.

ppb_vector/__init__.py
AstraLuma: I honestly haven't looked into why these lint rules exist.

.cirrus.yml
AstraLuma: I like specifically listing the major versions. We should also list the development head, though.

nicoonoclaste:
@AstraLuma Should I deal with the merge conflict for you?

.cirrus.yml
nicoonoclaste: I guess the Linux and Windows CI are fast enough the duplicated work shouldn't be an issue.

AstraLuma: well, over the course of the year, the dev version will diverge from the last release. Testing against it gives us a heads up if a future change is going to break something.

Extremely unlikely for `ppb-vector`, but we might as well.

AstraLuma:
Handled.

ppb_vector/__init__.py
AstraLuma: Ok, in Python there's two forms of chaining, explicit and implicit, which use different special attributes to distinguish them and different messages ("the above was the direct cause of ..." vs "while handling the above exception, another exception occurred").

This lint rule is to force wrapper exceptions to use the former path and message rather than the latter, which is strictly speaking a better error message.

.cirrus.yml
nicoonoclaste: Oh, I think we misunderstood one-another. I wasn't suggesting not to test against `rc-slim` or such, I was suggesting that we should refer to the latest release as just `3-silm` (which always points to the latest stable release, AFAIK) so if we forget to update the version number we are still testing against the latest stable anyhow.

nicoonoclaste (approved):
LGTM, remaining nit can be picked in a later PR.

PR #202, opened 2021-10-14
Support positional matching

AstraLuma:
New for 3.10.


AstraLuma:
Not sure how to handle testing for this, given that the preferred tests won't even compile on <3.10.

nicoonoclaste:
@AstraLuma I'd suggest having a `test_py3.10.py` or such, and [selectively ignoring it in `conftest.py`](https://docs.pytest.org/en/6.2.x/example/pythoncollection.html#customizing-test-collection) based on Python's version number.

AstraLuma:
ok, now i officially have no idea why pytest isn't collecting properly.

AstraLuma:
@nbraud @pathunstrom I think this is ready for review.

pathunstrom (approved):
I like the methodology for testing forward looking features when they don't break syntax.

nicoonoclaste (approved):
LGTM, except it broke `lint.sh` on Python versions older than 3.10 (not unlikely on contributors' dev environment)

I wrote a fix for that, please review it and if it's fine by you then `bors r=nbraud,pathunstrom` ?

ppb_vector/__init__.py
```diff
@@ -64,6 +64,8 @@ class Vector:
     x: float
     y: float
 
+    __match_args__ = ('x', 'y')
```
nicoonoclaste, line 67: Might be nice to add a comment pointing to the right documentation (in the PEP or otherwise) for that dunder ^^

AstraLuma:
Oh, I just fixed it by updating the containers. But I guess that helps with development environments.

AstraLuma:
bors r=nbraud,pathunstrom

PR #203, opened 2021-10-14
Use homebrew for python

AstraLuma:
This will hopefully cut 8~10min off the mac CI times.

Depends on #199.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -125,40 +125,24 @@ macOS_task:
   env:
     PATH: ${HOME}/.pyenv/shims:${PATH}
     matrix:
-      # Sadly, I have not found a way to reliably discover what versions are
-      # available without just asking in a Cirrus job.
-      - PYTHON: 3.7.9
-      - PYTHON: 3.8.6
-      - PYTHON: 3.9.0
+      - PYTHON: 3.7
+      - PYTHON: 3.8
+      - PYTHON: 3.9
+      - PYTHON: 3.10
   install_script:
     # Per the pyenv homebrew recommendations.
     # https://github.com/pyenv/pyenv/wiki#suggested-build-environment
-    - brew install openssl readline pyenv
-    - pyenv install ${PYTHON}
-    - pyenv global ${PYTHON}
-    - pyenv rehash
-    - pip install --upgrade-strategy eager -U pip wheel setuptools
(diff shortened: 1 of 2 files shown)


AstraLuma:
Hmph, homebrew is still compiling python, but it's still closer to 8min instead of 12min.

@pathunstrom you think this is worth doing?

.cirrus.yml
```diff
@@ -119,24 +121,24 @@ macOS_task:
...
-      - PYTHON: 3.7.7
-      - PYTHON: 3.8.6
-      - PYTHON: 3.9.0
+      - PYTHON: 3.6
+      - PYTHON: 3.7
+      - PYTHON: 3.8
+      - PYTHON: 3.9
+      - PYTHON: 3.10
   install_script:
     # Per the pyenv homebrew recommendations.
```
nicoonoclaste: Obsolete comment needs removing. ;3

nicoonoclaste:
> Hmph, homebrew is still compiling python, but it's still closer to 8min instead of 12min.

I'm no macOS expert, so I'd let Piper weigh in on the desirability of having `pyenv`, but I'd say a 1/3rd reduction in CI time is very appreciable, especially as macOS tests are usually the thing holding CI up: FreeBSD tests tend to finish earlier, and the containerized tests for Linux & Windows take negligible time in comparison.

(Also that's not workable until Homebrew gets recipes for Py3.10)

pathunstrom:
4 minutes is a long time in CI (especially with our community rate limits). If it's that much faster to build and that's our only option at the moment, let's do it.

AstraLuma:
It's apparently somewhat variable--it was running 8min/run before, but today it's 12+.

AstraLuma:
I still like this because we at least don't have to touch mac every time there's a patch release.

pathunstrom:
> I still like this because we at least don't have to touch mac every time there's a patch release.

Reasonable stance. We keep losing time to solving this with major releases, and I'm definitely up for reducing our costs there, too.

nicoonoclaste (approved):
LGTM

nicoonoclaste:
@AstraLuma Feel free to merge with `r=nbraud` if this is complete.  Not merging myself since I don't know heck about macOS.

AstraLuma:
bors r=nbraud

nicoonoclaste:
Rebased to deal with merge conflict.
bors r+

PR #206, opened 2021-10-25
Vector: Add a decompose method for basis change

nicoonoclaste:
### Changes

- Extend `tests.utils.angle_isclose` to accept arbitrary moduli (like comparing angles up to 180°, if one doesn't care about sign)
- Add `Vector.decompose` and relevant tests

### Rationale

`Vector.decompose` lets one decompose any vector into 2 orthogonal components, the first of which is colinear to the provided `basis` vector.

### Extensions

We could accept non-normalized vectors, and/or non-orthogonal basis.
I felt this was out-of-scope for a first go at this.


AstraLuma (approved):
bors r+

PR #207, opened 2021-10-25
tests.utils: Require optional args be named where ambiguous

nicoonoclaste:
### Changes

- `tests.utils.angle_isclose`: Accept arbitrary moduli
  Same as from #206, avoids merge conflict.

- In `tests.utils` functions with multiple optional arguments, make those kw-only:
  `angle_isclose`, `isclose`, and `lengths` are affected, no changes needed at call-sites.
  

### Motivation

When there is more than one optional argument, using them positionally can cause
confusion and higher likelyhood of bugs.

Those aren't public interfaces, this change isn't breaking.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -16,7 +16,7 @@ def floats(max_magnitude=1e75):
     return st.floats(min_value=-max_magnitude, max_value=max_magnitude)
 
 
-def lengths(min_value=0, max_value=1e75):
+def lengths(*, min_value=0, max_value=1e75):
     return st.floats(min_value=min_value, max_value=max_value)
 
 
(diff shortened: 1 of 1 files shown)


tests/utils.py
```diff
@@ -16,7 +16,7 @@ def floats(max_magnitude=1e75):
     return st.floats(min_value=-max_magnitude, max_value=max_magnitude)
 
 
-def lengths(min_value=0, max_value=1e75):
+def lengths(*, min_value=0, max_value=1e75):
```
nicoonoclaste, line 19: I wasn't so sure about that one, since `lengths(10, 10_000)` or such seems relatively-obvious, but I chose to err on the side of safety since we never use that construction anyhow.

AstraLuma (approved):
bors r+

PR #208, opened 2021-10-25
Vector.update: Require args to be named

nicoonoclaste:
### Motivation

`foo.update(bar)` is a lot more confusing than `foo.update(x=bar)`

### Caveats

This technically is a breaking change, though `Vector.update` is a niche API and I doubt anyone is writing
`foo.update(x, y)` instead of `Vector(x, y)` or such.

### DIFF
--- a/ppb_vector/__init__.py
+++ b/ppb_vector/__init__.py
@@ -149,7 +149,7 @@ class Vector:
     def __reduce__(self):
         return Vector, (self.x, self.y)
 
-    def update(self,
+    def update(self, *,
                x: Optional[SupportsFloat] = None,
                y: Optional[SupportsFloat] = None):
         """Return a new :py:class:`Vector` replacing specified fields with new values."""


nicoonoclaste:
More generally, is there a `flake8` plugin or such that could warn us when we do that... and hopefully help us avoid releasing public APIs with that kind of issue?  :sweat_smile:

AstraLuma (approved):
Approved, but need to decide on a release cadence for ppb-vector.

nicoonoclaste:
Yeah, I think the reasonable solution would be to bump the `ppb-vector` major version, which would also render moot the question of whether dropping Py3.6 is a major change (though it would be nice if we could come to a consensus regardless).

My understanding of `ppb`'s [policy] is that the new version of `ppb-vector` can simply go in the next `ppb` release (southern solstice 2021) as there is no removal or significant change in the API.

[policy]: https://ppb.readthedocs.io/en/stable/discussion/maintenance-schedules.html#deprecations-and-api-breakages

pathunstrom:
I will say, this matches the pattern we've been moving to in ppb as a whole, so big thumbs up from me.

nicoonoclaste:
Rebased to deal with merge conflict.

bors r=astraluma,pathunstrom

PR #209, opened 2021-10-26
Improve annotations and comments

nicoonoclaste:
- `VectorLike`
  * Nix obsoleted comment
  * Refine type definition for mappings
- Comment `Vector.__match_args__`
- Simplify references to the `typing` module
- Document deprecation of typing.{Sequence, Tuple, ...} from Python 3.9 onwards
- `lint.sh`: Avoid linting `.eggs/` and other directories

### DIFF
--- a/ppb_vector/__init__.py
+++ b/ppb_vector/__init__.py
@@ -1,3 +1,4 @@
+import sys
 import typing
 import warnings
 from collections.abc import Mapping, Sequence
@@ -12,12 +13,23 @@ __all__ = ('Vector',)
 __version__ = "1.0"
 
 
+# TODO: TypedDict is only available from the stdlib in Python 3.8 onwards
+if sys.version_info.minor >= 8:
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
+
+# Mappings convertable to Vector
+VectorLikeDict = TypedDict('VectorLikeDict', {'x': SupportsFloat, 'y': SupportsFloat})
+
+
 # Anything convertable to a Vector, including lists, tuples, and dicts
 VectorLike = Union[
     'Vector',
     Tuple[SupportsFloat, SupportsFloat],
     typing.Sequence[SupportsFloat],  # TODO: Length 2
-    typing.Mapping[str, SupportsFloat],  # TODO: Length 2, keys 'x', 'y'
+    VectorLikeDict,
 ]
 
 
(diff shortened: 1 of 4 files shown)


ppb_vector/__init__.py
```diff
@@ -259,13 +272,13 @@ def scale_by(self, scalar: typing.SupportsFloat) -> 'Vector':
...
     @typing.overload
-    def __mul__(self, other: typing.SupportsFloat) -> 'Vector': pass
+    def __mul__(self, other: SupportsFloat) -> 'Vector': pass
 
     def __mul__(self, other):
         """Perform a dot product or a scalar product, based on the parameter type.
 
         :param other: If ``other`` is a scalar (an instance of
-          :py:class:`typing.SupportsFloat`), return
+          :py:class:`SupportsFloat`), return
```
AstraLuma: According to [the docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects) sphinx does not follow python's rules for name resolution.

AstraLuma: (but it might work anyway with the exhaustive search.)

setup.py
```diff
@@ -16,6 +16,6 @@ def requirements(section=None):
 # See setup.cfg for the actual configuration.
 setup(
     # setup needs to be able to import the library, for attr: to work
-    setup_requires=['pytest-runner'],
+    setup_requires=requirements() + ['pytest-runner'],
```
AstraLuma, line 19: pretty sure that's an `install_requires`.... not sure why it lists `pytest-runner` here, too.

AstraLuma (changes requested):

ppb_vector/__init__.py
pathunstrom: You can remove the `typing` by prepending a tilde: :class:`~events.Idle`

nicoonoclaste: Wow, rude.  OK I'll change that

setup.py
nicoonoclaste: Mew?  THe comment above should clarify: `setup` needs to be able to import the library, which itself imports from `typing_extensions` on Py3.7.

Yopu are right though ths shloud abe a `install_require` too

AstraLuma: Oh, I'll have to look at this more closely then.

nicoonoclaste:
@AstraLuma Handled the requested changes and fixed the merge conflict.

AstraLuma (approved):

nicoonoclaste:
Feel free to merge my PRs when you approve them; if they aren't ready I'm make them be drafts.
bors r=astraluma

PR #210, opened 2021-10-26
Vector.decompose: Fix slightly-flaky test

nicoonoclaste:
- utils: Make default MAX_MAGNITUDE into a constant
  This way, we can refer to the default vector magnitude rather than make up another threshold.

- tests/decompose: Fix name of linearity test
  While I was at it ;3

- tests/decompose/linear: Limit length of λv vector
  This is the actual fix

### DIFF
--- a/tests/test_decompose.py
+++ b/tests/test_decompose.py
@@ -38,6 +38,8 @@ def test_decompose_angles(v: Vector, basis: Vector):
 
 @given(
     basis=units(), v=vectors(),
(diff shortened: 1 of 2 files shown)


AstraLuma (approved):
yes, i totally understand these test changes.

nicoonoclaste:
> yes, i totally understand these test changes.

Does the comment help?

AstraLuma:
Yeah, that did actually help.

bors r+

Issue #171, opened 2019-06-08
Add Projection


AstraLuma:
Add a projection method.

The simplest implementation seems to be `(a * b.normalize()) * b.normalize()`.

nbraud:
@astronouth7303 Is that supposed to be the projection of `a` on the surface normal to `b`?
'cause that wouldn't be it, I think.

Anyhow, yes, having a projection method seems reasonable to me.

astronouth7303:
No, that would be the projection of `a` onto the vector `b`, not its normal, as per [wikipedia](https://en.wikipedia.org/wiki/Vector_projection).

Although maybe defining a version that operates on the normal make sense? Matches reflect then?

nbraud:
@astronouth7303 Both make sense:
- if we think of projecting and reflecting on surfaces (which makes extra sense in higher dimensions), we definitely want a normal;
- if we think of projecting and reflecting on a line, defining it by a collinear vector also makes sense.

I guess we could have `project_by` (for the projection on a vector's normal) and `project_unto` (the vector's collinear line), and similarly `reflect_by` and `reflect_unto` (and deprecate `reflect`, as the name is ambiguous).

If that's something we agree on (and I'd be happy if we find better names), I'd be happy to make it happen.

astronouth7303:
Higher dimensions is something that's explicitly out of scope?

nbraud:
@astronouth7303 This was a parenthesis specifically because of that: “it's an API which is sensible in 2D, and is the main thing you would want in 3D or above”.  I don't think it's out-of-scope to provide an API that's consistent with what happens in not-2D geometry (lowering cognitive load on users).

astronouth7303:
@pathunstrom any perspective from the teaching/student side?

pathunstrom:
I'm going to ask around, but first thoughts:

`reflect` is there because it's an algorithm I used regularly for physics based games (like Arkanoid clones). I've _never_ needed reflect across a line. So my thought on that is `Vector.reflect` should be the physics based version, and something like `Vector.reflect_across` for the across a line version. (Also I'd prefer `reflect_onto` over `reflect_unto` but neither of those feel accurate to me.)

I need to dig more into projection, another thing I've never needed so need to understand better before I can comment.

astronouth7303:
I realized we might need it because @\eevee was discussing some physics they were doing and one of the steps involved projecting the momentum onto the ground. (They were dealing with momentum going over a hill causing the character to catch some air.)

[...]

PR #211, opened 2021-10-27
Add Vector.project method

nicoonoclaste:
That's the projection function requested in #171.

New tests were run under `--hypothesis-profile lots` so they shouldn't be flaky.

### DIFF
--- a/ppb_vector/__init__.py
+++ b/ppb_vector/__init__.py
@@ -573,5 +573,14 @@ class Vector:
 
         return self - (2 * (self * surface_normal) * surface_normal)
 
+    def project(self, direction: VectorLike) -> 'Vector':
+        """Project a vector on a line, defined by its direction.
+
+        :param direction: A :py:class:`Vector` or vector-like, that is non-null.
+        """
+        basis = Vector.normalize(Vector(direction))
+        p, _ = self.decompose(basis)
+        return p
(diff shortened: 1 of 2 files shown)


ppb_vector/__init__.py
```diff
@@ -573,5 +573,14 @@ def reflect(self, surface_normal: VectorLike) -> 'Vector':
 
         return self - (2 * (self * surface_normal) * surface_normal)
 
+    def project(self, direction: VectorLike) -> 'Vector':
+        """Project a vector on a line, defined by its direction.
+
+        :param direction: A :py:class:`Vector` or vector-like, that is non-null.
+        """
+        basis = Vector.normalize(Vector(direction))
```
AstraLuma: Odd phrasing. Couldn't this just be `Vector(direction).normalize()`?

nicoonoclaste: It totally should!  I had just written `Vector.normalize(direction)` first but that didn't accept vector-likes (and that seems legit)

nicoonoclaste: @AstraLuma Done  <3

nicoonoclaste:
*pokes @AstraLuma.*

AstraLuma (approved):
bors r+

PR #212, opened 2021-10-27
Add & use class-level constants

nicoonoclaste:
- Add class-level constants `Vector.u` and `v` for each axis.
- Replace homemade constants in `tests/utils` with those.
- Replace hardcoding of `(0, 0)` with `zero` in `Vector`.

### DIFF
--- a/ppb_vector/__init__.py
+++ b/ppb_vector/__init__.py
@@ -91,13 +91,13 @@ class Vector:
         The vector with no length.
         >>> assert not Vector.zero
 
-    .. py:attribute:: Vector.u
+    .. py:attribute:: Vector.x_unit
         :type: Vector
         :value: Vector(1.0, 0.0)
 
         The unit vector aligned with the x axis.
 
-    .. py:attribute:: Vector.v
+    .. py:attribute:: Vector.y_unit
         :type: Vector
         :value: Vector(0.0, 1.0)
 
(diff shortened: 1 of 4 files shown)


nicoonoclaste:
`u` and `v` are classical names for those 2 vectors in mathematics; I'm unsure
whether they are the best ones for our users so I asked a computer gfx friend.

AstraLuma:
I'm not immediately thrilled with the names. Why not "x axis" or similar?

AstraLuma:
Also, all three need sphinx-friendly inline docs.

nicoonoclaste:
> I'm not immediately thrilled with the names. Why not "x axis" or similar?

I had discarded `Vector.x` because of the name collision, but yes `x_axis` — or perhaps `x_unit` — would probably be friendlier for neophytes.

nicoonoclaste:
> Also, all three need sphinx-friendly inline docs.

Definitely. I had just skipped it for the time being because I couldn't find how to attach them a docstring, but apparently that's not a thing so I'm making do with the class-level docstring and a bunch of rST markup  :sweat_smile:

nicoonoclaste:
*grumbles about VS Code's search&replace and wonders why `mypy` didn't catch this.*

nicoonoclaste:
Spurious CI failure (fixed in `canon`)

AstraLuma:
Use `#:` per https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html

nicoonoclaste:
> Use `#:` per https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html

Is that the right thing for class-level variables (as opposed to object-level members) ?

AstraLuma:
Not sure autodoc makes a distinction?

pathunstrom:
Autodoc does not make the distinction.

`#:` is the only way to have autodoc include arbitrary documentation that I've found.

nicoonoclaste:
> Use `#:` per https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html

Done.

AstraLuma (approved):
bors r+

AstraLuma:
bors retry

PR #213, opened 2021-10-27
Simplify CI definitions

nicoonoclaste:
- [x] Simplify Windows image names
- [x] Automatically test against latest CPython
- [x] Automatically test against latest ~~& development~~ PyPy
- [x] Add PyPy 3.8


pathunstrom:
Just because I'm curious: is there an actual semantic difference between the servercore version and just the python versions?

nicoonoclaste:
> Just because I'm curious: is there an actual semantic difference between the servercore version and just the python versions?

No difference whatsoever: both are shared tags that refers images for that particular CPython version and various Windows version; the one without `-windowsservercore` also refers to Linux images but that doesn't affect our Windows builds  ^^

AstraLuma (approved):
bors r+

PR #214, opened 2021-10-28
Parallelise the testsuite in dev (as possible) and in CI

nicoonoclaste:
### Changes

- [x] `test.sh` checks whether `pytest-xdist` is available, and if so runs the test suite in parallel
- [x] `pytest-xdist` is installed in CI's test environments
- [x] Introduce a new `requirements-ci.txt` for those environments, which cascades to `requirements-tests.txt`

### Motivation

Our test suite is getting slower, as it grows deeper and our API grows wider.  While unsurprising, this is an issue in 2 ways:

1. During (local) development, `test.sh` taking 8s is enough of a disruption that developers may subconsciously skip it, or specifically run a subset of the test suite, potentially missing issues which are only caught in CI once pushed, and getting much slower feedback.
2. In CI, `test.sh` takes about 1'30" (on Linux) which is long-enough one may skip to a different task etc., introducing further delays in the feedback loop, and more cognitive overhead from repeated context switching.  Ideally, I'd like our CI to complete within 10s of seconds (at least on PRs themselves, Bors may take longer as it's async)

Parallelising the test suite addresses both usecases:
1. `test.sh` got 4× faster in local development — 8s to 2s — fast-enough I could consider running it in a commit hook or such.
2. In the CI profile, the speedup was close to 8× — from 1'23" to 11" — which achieves my goal of getting feedback in tens of seconds.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -89,7 +89,6 @@ FreeBSD_task:
 task:
   skip: $CIRRUS_BRANCH =~ '.*\.tmp'
   env:
-    PYTEST_CPUS: 8
     matrix:
       - IMAGE: python:3.7-slim
       - IMAGE: python:3.8-slim
@@ -103,8 +102,8 @@ task:
   name: "Linux $IMAGE"
   allow_failures: $IMAGE =~ '.*-rc-.*'
   container:
-    cpu: $PYTEST_CPUS
     image: $IMAGE
+    greedy: true
 
   install_script:
     - pip install --upgrade-strategy eager -U -r requirements-ci.txt
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -89,6 +89,7 @@ FreeBSD_task:
 task:
   skip: $CIRRUS_BRANCH =~ '.*\.tmp'
   env:
+    PYTEST_CPUS: 8
     matrix:
       - IMAGE: python:3.7-slim
       - IMAGE: python:3.8-slim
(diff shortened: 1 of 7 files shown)


nicoonoclaste:
Apparently, `pytest-xdist` is chocking on `@pytest.mark.parameterize`d tests which seem to be instanciated in non-deterministic order; I suspect the source of non-determinism might be the traversal order of `frozenset`s defined in `utils.py`.

nicoonoclaste:
CI failures seems to be down to Cirrus' VMs being unusually slow, maybe due to noisy neighbours or such.

nicoonoclaste:
bors try

AstraLuma (approved):
That's some high standards. While waiting for mac is the much bigger problem, this is an improvement with minimal maintenance burden.

bors r+

nicoonoclaste:
bors r=astraluma

nicoonoclaste:
*grumbles*  Of course, requesting 8 CPUs causes the task to take forever before being scheduled  >_>'

On the plus side, I found the root cause of the flakyness issue: `pytest-xdist` was grabbing 30 threads, one for each logical core of the host, even though it was running in a container limited to 2 cores  ^^'

nicoonoclaste:
CPU usage literally off the charts ^^'

2021-11-02_off_the_charts

nicoonoclaste:
Apparently, spinning up 30 PyPy processes does cause jobs to OoM, though  >_>'

nicoonoclaste:
@AstraLuma Care to re-review?  I feel like there was enough ~~flailing around~~ incremental fixing to warrant that  ^^

AstraLuma (approved):
lgtm

bors r+

PR #215, opened 2021-11-02
CI/Lint: Simplify task requirements w/ cascading

nicoonoclaste:
I think it makes sense to have a single file specifying what we need for linting,
rather than what we need in addition of the test dependencies.

This uses the same feature of requirements files that #213 relies on, so nothing
new or more complex.


AstraLuma (approved):
bors r+

PR #217, opened 2021-12-12
Selectively run CI environments outside Bors

nicoonoclaste:
### Motivation

CI is currently slow, both in providing feedback on PRs, and in clearing through
Bors requests.

The issues compound when a PR branch is updated, causing a flurry of jobs to be
executed, right before Bors is instructed to merge, as both are competing for
jobs to be scheduled, under a shared concurrency limit.

### Main changes

- Skip “expensive” environments (FreeBSD, macOS) on immediate PR feedback.  
  They still run under Bors, enforving tests pass there, but we avoid consuming
  scarce, free CI time.
  
- Use PPB's compute credits for test tasks Bors is waiting on.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -37,6 +37,7 @@ pep517_task:
     - python3 -m pep517.check .
 
 upload_task:
+  skip: $CIRRUS_BRANCH =~ '.*\.tmp'
   # only_if: $CIRRUS_RELEASE != ""
   env:
     TWINE_REPOSITORY_URL: "https://test.pypi.org/legacy/"
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -63,6 +63,7 @@ upload_task:
 
 FreeBSD_task:
   only_if: $BRANCH == 'staging' || $BRANCH == 'trying'
+  use_compute_credits: yes
   freebsd_instance:
     image_family: freebsd-13-0
   env:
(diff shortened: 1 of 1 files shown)


AstraLuma:
We've also been discussing just getting our own mac, so we have guaranteed CI capacity there.

nicoonoclaste:
> We've also been discussing just getting our own mac, so we have guaranteed CI capacity there.

That seems orthogonal to what this PR does?  Even assuming the macOS CI jobs are immediately scheduled, they take several minutes more than the Linux and Windows jobs.

Moreover, we have multiple macOS jobs, so they'd probably have to execute sequentially, even assuming no other PR is submitted concurrently.

nicoonoclaste:
@AstraLuma Looks like Cirrus must be using a custom YAML parser that recognises `true` but not `yes`  >___>'

nicoonoclaste:
bors retry

AstraLuma:
Not exactly

https://cirrus-ci.com/build/6237841507221504

nicoonoclaste:
> Not exactly
> 
> https://cirrus-ci.com/build/6237841507221504

Ayup, but previously the thing didn't even go that far (on this PR) only got an error message about failing to evaluate the config  :3

Please keep us posted if/when Cirrus' support replies though  <3

pathunstrom:
We got an update this morning: Apparently blocking application builds is a security thing and they'll lift the restriction once they fix the whole thing.

PR #221, opened 2022-11-12
Remove spurious setup requirements, and setup.py

nicoonoclaste:
- Drop spurious setup requirements
  `setuptools` can now deal with `attr:` without needing to import the package.
- Move dependency options to `setup.cfg`
  There was nothing non-trivial left in `setup.py`, so I removed it too  ;3


nicoonoclaste:
The CI failure on Windows/Py3.9 is spurious (Cirrus' VM didn't cleanly come up apparently)

pathunstrom:
bors retry

